PR #158, opened 2026-08-14
feat(mobile): login and dashboard UX pass

dmadan86:
Login and dashboard UX pass, verified live on an Android emulator.

## Dashboard
- **Category filter strip** under "Your groups" — icon-over-label chips (All + one per group type you actually have), filters the group list. Only shows when there's more than one type; a stale filter snaps back to All.
- **Removed the Quick actions deck.**

## Sign-in
- **Tappable country-code picker** on the OTP tab: flag + `+NN` chip → searchable country sheet. The number field holds only local digits; the wire format sent to Supabase is E.164.
- Dial code **defaults to the device Region** (per expo-localization, which reads the Region setting, not the display language), falling back to India only when the region is unknown. `deviceCountry()` now walks the full locale list for robustness.
- **Sign-in ↔ sign-up toggle moved above the fields** (the Android keyboard was hiding it below the submit) and reworded to an explicit action ("Already have an account? Sign in").
- **Generic placeholder email** (`alex@example.com`); phone-code hints follow the device/selected country instead of a hardcoded `+91`, on both sign-in and the account add-contact screen.
- **Cold-start flash fix**: the auth gate no longer paints the dashboard for a frame before redirecting a signed-out user to sign-in — it holds the spinner until route and session agree.

## Core
- `dialingCodeForCountry()` added to `money/region.ts`.

## Member screen (admin discoverability)
- Role card shown disabled-with-reason for ghosts; admin badges on the roster and balances rows.

## Verification
Driven on the emulator: cold start → onboarding → welcome (no dashboard flash) → sign-up form (top sign-in link + generic placeholder) → OTP tab picker (🇺🇸 +1 → tapped India → 🇮🇳 +91). Gate green: typecheck, prettier, eslint.

### DIFF
--- a/apps/mobile/src/app/(tabs)/index.tsx
+++ b/apps/mobile/src/app/(tabs)/index.tsx
@@ -15,10 +15,8 @@ import {
   Screen,
   SectionHeader,
   Text,
-  TintCard,
   useTabBarClearance,
   useTheme,
-  type TintName,
 } from '@baaki/ui';
 
 import { useGroups, useHomeSummary } from '@/data/hooks';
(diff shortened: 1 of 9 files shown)


dmadan86:
Rebased onto latest `main` (atop the security-hardening merge #159). Gates green: typecheck (7 workspaces), prettier, tests (core 517, mobile 206, web 24). No conflicts. Ready for review.

PR #227, opened 2026-08-16
perf(mobile): cross-fade tab scenes instead of shift

dmadan86:
## Why

Tab switches felt sluggish. The bottom-tab scene animation was `shift`, which slides each scene sideways *and* fades it. The horizontal translate is the heaviest work in the tab-switch path; on a busy destination (the dashboard — hero carousel + queries mounting) it drops frames, so the switch reads as slow.

## Change

`(tabs)/_layout.tsx`: scene `animation` `shift` → `fade`. A soft cross-fade that still honours the motion switch at a fraction of the cost. `none` when motion is off, unchanged.

## Note on scope

Reported on a **dev apk + Metro** build, where dev-mode Hermes + Metro instrumentation inflate transition cost well beyond release. This trims the one JS-thread-heavy animation in the tab path, which helps both dev feel and real low-end devices — but the true measure is a release build.

typecheck + lint green. No test pinned `shift`.

### DIFF
--- a/apps/mobile/src/app/(tabs)/_layout.tsx
+++ b/apps/mobile/src/app/(tabs)/_layout.tsx
@@ -16,9 +16,13 @@ export default function TabsLayout() {
         headerShown: false,
         // The root bar draws the navigation; this one would be a second copy.
         tabBarStyle: { display: 'none' },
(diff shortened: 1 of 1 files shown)


dmadan86:
Superseded: removing screen transition animation entirely (fade → none) instead of lightening it.

PR #276, opened 2026-08-19
fix(mobile): give the captures inbox an activity-style loading skeleton and layout

dmadan86:
## Summary
- The captures inbox rendered `rows = captures.data ?? []` and only checked `rows.length === 0`, so while `captures.isLoading` was true it showed the "nothing yet" empty state instead of a loading placeholder, then popped in cards once the local mirror hydrated.
- Fixed by mirroring `ActivityScreen`'s pattern exactly: one persistent `ScrollView` (with pull-to-refresh) wraps all three states — `isLoading` → skeleton, loaded-and-empty → `EmptyState`, loaded-with-rows → content — instead of a separate `View`/`ScrollView` per branch (which also meant pull-to-refresh didn't work in the old empty state).
- Added a new `InboxSkeleton` to `Skeletons.tsx`, shaped to the real capture `Card` down to the pixel (46px thumbnail, subheading/caption/micro text lines, the `sm` Assign button and delete `IconButton` footprint) — same philosophy as the existing `FeedSkeleton`/`PeopleSkeleton`.
- Brought the inbox's layout closer to the activity feed by day-grouping capture cards under the same `dayHeading` headings the activity tab uses. Lifted `groupByDay` out of `activity.tsx` into `data/activity.ts` so both feeds share one implementation.
- Kept cards (not the activity timeline) since captures carry photo thumbnails, category badges, and per-row assign/delete actions that a timeline row can't hold — the skeleton and day-grouping now do the "one system" work instead.
- All existing inbox behavior (assign-to-group sheet, delete confirm, capture CTA, offline "not synced yet" hint) is unchanged.

## Test plan
- [x] `pnpm format`
- [x] `pnpm lint`
- [x] `pnpm --filter @waves/mobile typecheck` (full `pnpm typecheck` fails only in `packages/db`, a pre-existing environment gap in this worktree — missing `packages/db/.env` with `DIRECT_URL` — unrelated to this change)
- [ ] Manual: open the inbox cold and confirm the skeleton shows before cards, and that a truly empty inbox shows the empty state (not skeleton) once hydrated

### DIFF
--- a/apps/mobile/src/app/(tabs)/activity.tsx
+++ b/apps/mobile/src/app/(tabs)/activity.tsx
@@ -16,30 +16,13 @@ import {
   useTheme,
 } from '@waves/ui';
 
-import { dayHeading, dayKey, describeActivity, parseMoney, verbIcon } from '@/data/activity';
+import { dayHeading, describeActivity, groupByDay, parseMoney, verbIcon } from '@/data/activity';
 import { fetchRecentActivity } from '@/data/api';
 import { FeedSkeleton } from '@/components/Skeletons';
 import { useNotifications } from '@/data/hooks';
 import { useStrings } from '@/i18n';
 import { useAuth } from '@/lib/auth';
 import { usePullRefresh } from '@/lib/pullRefresh';
-
-/**
- * The feed cut into calendar days, newest first, order otherwise untouched — the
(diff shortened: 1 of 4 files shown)


dmadan86:
@coderabbitai help

dmadan86:
@coderabbitai resume

PR #356, opened 2026-08-22
feat(mobile): clone a group — seed a new group from an existing one, notify the people already on Waves

dmadan86:
## What

Make a new group from one you already have.

- **Two entry points** — *Duplicate group* inside a group's settings, and *Start from an existing group* on the New Group screen, which opens a source picker (favourites first).
- **The clone seeds the New Group form** from the source: name (`X copy`), icon (emoji), kind, trip dates, budget, simplify — all editable.
- **People are carried as the same removable chips** a fresh group uses, so you drop anyone before the group is written. Each carries whatever address the source row held.
- **Photo is NOT copied** — a group photo's storage object is scoped to the source group's path (the read policy checks membership of the group in the path), so a copied reference would render only for source-members. The clone keeps the emoji; a photo is set later in settings, uploaded under the new group's own path and paid-gated. (Supersedes the earlier "copy the photo" intent once the storage-scoping constraint was found.)
- **Favourites** are a device-local star (no migration, no sync): floats a group to the top of the picker; toggled from settings or the picker.

## Notify the people already on Waves

`baaki_add_ghost_member` now, **after a genuine new add**, matches the added email/number to an `auth.users` account and — if it isn't the caller and isn't already a member — writes a `group_added` notification, delivered by the existing **notify-fanout** (push + inbox, rendered in the reader's language).

- The two idempotent paths above the insert return **before** the notify; a per-`(group, person)` dedupe key is the second belt → a retried offline mutation or a re-add never buzzes twice.
- The `auth.users` read is **guarded on the table existing**, so the DB tests on bare Postgres stay green.

## Layers touched
- **DB** — migration `20260822120000_group_add_notify` (RPC edit only; no table change → no drift). **Applied to prod.**
- **core** — `NotificationKind.GroupAdded` + copy en/ta/hi/ar (total record).
- **mobile** — `clone-group.tsx` picker, `lib/favorites.ts` (pure store + hook), New Group `?from=` seed, settings Duplicate + star, route registration, i18n.

## Tests
- **DB** `group-add-notify.test.ts` — the notify half via its refusals: match-by-email/phone tapped once, self-add skipped, existing-member skipped, re-add deduped, plain-ghost + no-match silent. Stubs `auth.users` with `phone`/`deleted_at`. **Ran green against local Postgres: 7/7, and 475/475 across the full db suite.**
- **core** — explicit `group_added` render + the existing all-languages parity test. **572 green.**
- **mobile** — `favorites` store: toggle, independence, empty-id no-op, subscriber fan-out, survives cold reload. **304 green.**
- **e2e** — `e2e/clone-group.yaml` (Maestro): settings → Duplicate → prefilled → drop a chip → Create → lands on the copy, original stays. (Needs an emulator.)

## Deviation to fold into the spec
- **TDR §7.1**: new notification kind `group_added` — and adding a contact who is already a Waves user now taps them in **any** group, not only on clone.

## Gates
✅ prettier · ✅ `tsc` (core + mobile) · ✅ eslint `--max-warnings 0` · ✅ core 572 · ✅ mobile 304 · ✅ db 475

## Summary by CodeRabbit

- **New Features**
  - Added the ability to duplicate existing groups, including editable settings, dates, budget, emoji, and eligible members.
  - Added group favorites with starring and un-starring controls.
  - Added a group selection screen with favorite prioritization, metadata, member counts, and empty states.
  - Added notifications when a new group member matches an existing account.
  - Added localized cloning support in English, Tamil, Hindi, and Arabic.

- **Improvements**
  - Added duplicate and favorite actions to group settings.
  - Added accessible labels and modal navigation for the cloning flow.

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -516,6 +516,7 @@ function AuthGate() {
           <Stack.Screen name="guest-welcome" />
           <Stack.Screen name="(tabs)" options={{ animation: 'none' }} />
           <Stack.Screen name="new-group" options={modal} />
+          <Stack.Screen name="clone-group" options={modal} />
           {paywallEnabled ? <Stack.Screen name="paywall" options={modal} /> : null}
           <Stack.Screen name="capture" options={modal} />
           <Stack.Screen name="captures" />
--- a/apps/mobile/src/app/clone-group.tsx
+++ b/apps/mobile/src/app/clone-group.tsx
@@ -1,0 +1,137 @@
+import Ionicons from '@expo/vector-icons/Ionicons';
+import { router } from 'expo-router';
+import { Pressable, ScrollView, View } from 'react-native';
+
+import { Avatar, IconButton, iconSize, Row, Screen, Text, tintForKey, useTheme } from '@waves/ui';
+
+import { useGroups, useHomeSummary } from '@/data/hooks';
+import { groupLabel } from '@/data/types';
+import { useAuth } from '@/lib/auth';
+import { useFavorites } from '@/lib/favorites';
+import { fill, plural, useStrings } from '@/i18n';
+
+/**
+ * Pick a group to start a new one from (the clone flow's front door).
+ *
+ * Everything about the chosen group — its name, icon, kind, dates, budget and
+ * who is in it — is carried into the New Group screen as a starting point, where
+ * it can all be changed and people dropped before anything is written. So this
+ * screen only has to answer one question, "which one", and it answers it the way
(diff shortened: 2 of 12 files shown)


dmadan86:
Addressed CodeRabbit review in the latest commit: RTL-aware chevrons at both clone entry points, a hydration-race guard on the favourites store (with a race test), and group-scoped notification counts in the DB test.

Not applying the "set `group_members.profile_id` for a matched account" suggestion: auto-linking a contact match to its account would add someone to a group without their consent. Per ADR-006 a contact-add is deliberately a *ghost* that the real account claims. The notification-landing experience for an already-registered match is a separate follow-up.

PR #414, opened 2026-08-24
fix(invite): show the QR by default — mint on open, share stays a separate button

dmadan86:
## Reported
Opening the group's QR/invite screen showed a **"Create link"** button first — you
had to tap it to generate before any QR appeared. Expectation: the QR should just
be there; creating/sharing the link is a separate action.

## Fix
Mint the invite **once on open** (a ref-guarded effect) so the QR renders straight
away. While the mint is in flight, a spinner card holds the QR's place. The
create-link button now only appears as a **retry** if the auto-mint fails. Share
and Copy are unchanged — separate buttons that reuse the same link the QR encodes.

## Note on interpretation
The QR encodes the minted invite token (so scanning it and tapping the shared link
join the *same* invite). "Show it by default" therefore means the mint happens
automatically/silently on open, not that the QR is tokenless. If you actually want
a **tokenless** group QR (no server invite created until you tap Share), that's a
backend change (a join-by-group path) — say the word and I'll scope it.

## Scope
One file, mobile-only. typecheck + lint clean. Hot-reloadable, nothing to deploy.

## Summary by CodeRabbit

* **New Features**
  * Invite links are now generated automatically when the invite screen opens.
  * Added a loading state while the invite link is being created.
  * Users can retry link creation if the initial attempt fails.

### DIFF
--- a/apps/mobile/src/app/group/[id]/invite.tsx
+++ b/apps/mobile/src/app/group/[id]/invite.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import Ionicons from '@expo/vector-icons/Ionicons';
 import { decode as decodeBase64 } from 'base64-arraybuffer';
 import * as Clipboard from 'expo-clipboard';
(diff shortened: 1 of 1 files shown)


dmadan86:
Superseded by the industry-standard durable group join link (stored groups.join_code + join-by-code + admin reset), which replaces the auto-mint-per-visit approach here.

PR #458, opened 2026-08-25
fix(invite): idempotent invite-use consumption for ghost-claims

dmadan86:
## The hole

`invite-accept` called `baaki_consume_invite` **before** the `body.claimMemberId` branch, so a ghost-claim burned a `max_uses` slot the instant it was tapped — every tap. A pending claimant is not a member, so `baaki_request_member_claim` returned `{ pending: true, alreadyPending: true }` on repeats while the edge function had **already re-run the consume each time**. Result: one person could empty a still-valid link by re-POSTing the same claim (or N bogus claims on N ghosts), `max_uses` burned with nobody joining, organizer-facing availability degraded.

Verified against current code before acting; the repo already carried `packages/db/test/invite-claim-use-count.test.ts` locking in the buggy behavior with `TODO(product)` markers — those are now replaced with the fixed behavior.

## Semantic chosen — option 2: consume at most once per `(invite, profile, member)`, idempotently

Consumption moves **out** of `invite-accept`'s claim path and **into** `baaki_request_member_claim`, positioned so it fires exactly once for a genuinely-new claim and never for a repeat or a doomed one:

- `NOT_CLAIMABLE` / `ALREADY_CLAIMED` / `ALREADY_A_MEMBER` → no use spent
- already-pending (same person asking again) → no use spent; the existing claim is the answer → **closes the repeat/bogus exhaustion abuse**
- a brand-new, valid claim → one use spent, atomically, in the same transaction as the `INSERT`. If the link is spent/invalid by then → refused (`INVITE_INVALID`), nothing written.

The consume is the existing `baaki_consume_invite` atomic conditional `UPDATE`, run under the member's `FOR UPDATE` lock in the same transaction as the claim row, so two new claims racing for a link's **last** slot (or a claim racing a direct join) serialise on the invite row — exactly one wins. A **direct (non-claim) join is unchanged**: `invite-accept` still calls `baaki_consume_invite` once, right before inserting the membership → a normal join still spends exactly one use.

### Why not "consume on approval" (options 1/3)?
That needs to carry the invite id to `baaki_decide_member_claim`, i.e. a new column on `member_claims` — which **is** a Prisma model (`MemberClaim`), so it would introduce Prisma drift. Reserving idempotently at request time closes the reported abuse with a **functions-only** migration and no schema drift.

## Changes

- **Migration** `20260825240000_invite_claim_idempotent_consume` — `DROP`s the old 4-arg `baaki_request_member_claim` (signature change → avoid a leftover ambiguous overload) and `CREATE`s a 5-arg version adding `p_invite_id uuid DEFAULT NULL`; consumes one use only for a new valid claim. `NULL` invite id → no consume (internal callers / existing 4-arg tests).
- **Edge** `supabase/functions/invite-accept/index.ts` — removed the pre-branch consume; claim path passes `p_invite_id: invite.id` and maps `INVITE_INVALID` → 404; direct-join path now consumes right before the membership insert.
- **Tests** `packages/db/test/invite-claim-use-count.test.ts` — rewritten from the `TODO(product)` baseline.

## Tests (7 pass; full requested matrix)

- repeated identical ghost-claim does **not** increment `use_count` (5 taps → `use_count` stays 1)
- bogus claims (`ALREADY_CLAIMED` / `NOT_CLAIMABLE` / `ALREADY_A_MEMBER`) spend nothing → cannot exhaust `max_uses`
- claim on an already-spent link → `INVITE_INVALID`, no pending row written
- **concurrent** redemption of the last slot via two live connections → exactly one `ok`, one `INVITE_INVALID`, `use_count = 1`, one pending claim
- pending-claim response stays idempotent (`already_pending`, same `claim_id`)
- normal (non-claim) accept still consumes exactly one use (regression guard)
- durable join link (huge `max_uses`) still lets a claim through and stays live
- existing `memberClaims.test.ts` unchanged and green (4-arg calls resolve via the default)

## Gates
`pnpm format` clean · `pnpm lint` all projects done · `pnpm edge:build` ok · `pnpm --filter @waves/db test invite-claim-use-count memberClaims` → all green. (4 unrelated attachment-image test files fail identically on baseline — local test-DB storage-trigger setup gap, not this change.)

## Deploy note
Owner deploys: **migrate deploy** of the new migration **+ edge deploy of `invite-accept`** (the RPC signature and the edge ordering must land together). Not deployed here.

## Summary by CodeRabbit

- **New Features**
  - Invite uses are now reserved atomically when a new membership request is created.
  - Repeated, invalid, or rejected requests no longer consume invite uses.
  - Exhausted invites are rejected with a clear error.
  - Direct joins continue to consume invite uses correctly while preserving existing-member and guest-limit behavior.
  - Concurrent requests allow only one successful invite-use reservation.

### DIFF
--- a/packages/db/prisma/migrations/20260825240000_invite_claim_idempotent_consume/migration.sql
+++ b/packages/db/prisma/migrations/20260825240000_invite_claim_idempotent_consume/migration.sql
@@ -1,0 +1,165 @@
+-- A pending ghost-claim must not burn an invite use it can never give back.
+--
+-- The hole (invite-accept): `baaki_consume_invite` ran BEFORE the claim branch,
+-- so the moment someone tapped "I'm Ravi" a `max_uses` slot was spent — whether
+-- or not an admin ever agreed, and, worse, EVERY time they tapped. A pending
+-- claimant is not a member, so `baaki_request_member_claim` happily returned
+-- `already_pending` on the second, third, hundredth attempt while the caller had
+-- already re-run the consume each time. A valid link capped at N uses could be
+-- emptied by one person re-POSTing the same claim N times, or by N bogus claims
+-- on N ghosts, with nobody ever joining. The organizer's link looks "full"; no
+-- one is in the group.
+--
+-- The fix (semantic #2 from the brief — "consume at most once per
+-- (invite, profile, member), idempotently"): consumption moves OUT of the edge
+-- function's claim path and INTO this function, placed so it fires exactly once,
+-- for exactly one genuinely-new pending claim, and never for a repeat or a
+-- doomed request:
+--
+--   * NOT_CLAIMABLE / ALREADY_CLAIMED / ALREADY_A_MEMBER  -> no use spent.
+--   * already-pending (the same person asking again)      -> no use spent;
+--     the existing claim is the answer, so the link is not touched. This is the
+--     idempotency that closes the repeat-exhaustion abuse.
+--   * a brand-new, valid claim                            -> one use spent,
+--     atomically, in the same transaction as the INSERT. If the link is spent or
+--     invalid by then, the claim is refused and nothing is written.
+--
+-- Because the whole function is one transaction and the member row is held
+-- `FOR UPDATE`, two concurrent requests for the same place cannot both file a new
+-- claim — the loser sees the winner's pending row and returns already_pending
+-- without consuming. And `baaki_consume_invite` itself is a single atomic
+-- conditional UPDATE, so two DIFFERENT new claims racing for a link's LAST slot
+-- (or a claim racing a direct join) serialise on the invite row: exactly one wins.
+--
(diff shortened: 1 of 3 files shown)


dmadan86:
Addressed the review in c2935a6:

- **CI red (typecheck + lint):** `test/invite-claim-use-count.test.ts:280` `losers[0].reason` tripped TS2532 (possibly undefined) under strict null checks. Guarded as `losers[0]?.reason`. (The known attachment-image `OBJECT_NOT_COMMITTED` failures are unrelated and fixed by #457.)
- **Nitpick 1 (pendingClaimCount):** now filters `status = 'pending'` so the helper matches its name.
- **Nitpick 2 (p_invite_id => NULL coverage):** added a test that files a new claim with a null invite id and asserts no invite use is spent (the internal / 4-arg-caller path).

Gates green: db tests 8/8, `tsc --noEmit` clean, `pnpm edge:build` ok, format + lint clean.

PR #487, opened 2026-08-26
test(core): cover multi-user split performance

dmadan86:
## Summary
- add multi-user split scenarios for a paid expense across equal, percentage, and share splits
- validate 4-user exact outcomes and 50-user ledger/balance invariants
- add a 50-user performance regression benchmark over 3,000 split computations

## Validation
- pnpm --filter @waves/core test -- split.multiuser.performance.test.ts
- pnpm --filter @waves/core typecheck

## Summary by CodeRabbit

* **Tests**
  * Added coverage for equal, percentage-based, and weighted expense splits involving up to 50 participants.
  * Added validation for share totals, participant coverage, net balances, and proportional rounding.
  * Added a 3,000-operation throughput test for 50-participant splits, verifying that each allocation preserves the original expense amount.


dmadan86 (commented):
@
Code review — `test(core): cover multi-user split performance`

Solid coverage: the 4-user exact-outcome cases across equal/percent/shares, the 50-user conservation + proportional-bounds invariants, and the ledger reconciliation helper are all well-targeted at what actually breaks (money not summing to the total).

One concern worth addressing before merge:

- **The wall-clock perf gate is CI-flaky.** `expect(averageMs).toBeLessThan(2)` in `split performance benchmark` asserts on absolute elapsed time. Shared CI runners vary wildly under load, so this will fail spuriously on a busy worker even when nothing regressed — the same class of flakiness the monkey suite already hit. Recommend one of: drop the timing `expect` and keep the loop purely as a correctness/throughput exercise (every `sumShares === amount` is the real guard); or make it log-only; or if a guard is genuinely wanted, raise the budget to something no realistic runner trips (e.g. an order of magnitude) and mark it clearly as a smoke ceiling, not a benchmark.

Nits (optional):
- `max`/`min` helpers duplicate a tiny bit of logic — fine as-is, but `computeShares` bounds could lean on the existing core helpers if any exist.
- The correctness assertions alone (conservation + bounds) would justify the file even without the timed block.

Non-blocking otherwise — the correctness half is good.
@

PR #575, opened 2026-08-28
feat(captures): reframe Drafts as a Saved-for-later inbox

dmadan86:
## Why
UX roast of the Drafts screen: the feature is strong (save now, assign later) but the UI asked people to learn the system model *capture → draft → assign → expense*. Users only think **"I paid for something; save it now, sort it later."** This reframes the screen around that.

## What
**Rename + softer wording (en/ta/hi/ar)**
- "Drafts" → **Saved for later**; dashboard card "Unassigned" → "Saved for later / *N* waiting to be added". Avoids the removed notification-Inbox name collision (#565).
- Empty state, form headers, save button all drop the "draft" framing.

**Row redesign (#3+#4 of the roast)**
- The always-red trash + pencil on every row fold behind one neutral **⋯** sheet: **Add to group** (primary, labelled) / Edit / Delete. Card tap still assigns — now it's the one obvious action, and delete no longer rides a mis-tap away from it. Batch cards get the same ⋯.

**Guidance**
- Header shows the waiting count under the title.
- Voice batches carry a one-line hint: *assign together, or open to handle each*.
- Assign sheet copy trimmed to one line.

**New-group carry-through (#8 — a real bug)**
- "New group" from the picker used to `router.push('/new-group')` and **drop** the pending assignment, so "Create one and add this to it" was a lie. Now the group is made, then the capture drops straight into its add-expense form. Shared `assignCaptureHref` keeps the picker and new-group hand-off identical.

## Deliberately skipped
- Header `+` — starting a capture lives on the dashboard FAB (existing decision); a second entry point is marginal.
- Status buckets ("Needs amount") — dead by construction (save is disabled at amount 0).
- "More details" fold on the capture form — it's already the lean version of add-expense.

## Testing
- `pnpm --filter @waves/mobile typecheck` ✅
- `pnpm --filter @waves/mobile lint` ✅
- `pnpm --filter @waves/mobile test` — 657 passed ✅
- Not device-tested (no device/emulator here); the ⋯ and assign use real `Modal`s, not in-tree overlays, so they float above the root tab bar like the existing assign sheet.

## Summary by CodeRabbit

* **New Features**
  * Added overflow actions for individual saved expenses and batches, including assign, edit, and delete.
  * Added direct assignment to groups with expense details prefilled.
  * Preserved selected expenses when creating a new group for assignment.
  * Added assignment guidance, pending counts, and clearer amount alignment for batch entries.

* **Localization**
  * Updated capture terminology to “expenses saved for later” across supported languages.
  * Added translated labels for batch guidance and additional actions.


dmadan86:
Addressed CodeRabbit review:

- **A11y (Major) — nested focusable overflow:** Fixed. The `⋯` (and the batch expander's `⋯`) are taken out of the accessibility tree and surfaced as a custom `more` `accessibilityAction` on the row's Pressable, so assign/expand stays the default activate and edit/delete/delete-batch are reachable via the action rotor. Sighted taps unchanged. `⋯`+chevron wrapped together to keep the amount column aligned.
- **RTL amount alignment:** No physical-edge code was used — the amount column is `alignItems: 'flex-end'`, which is layout-direction-relative in RN (resolves to the trailing edge in both LTR and RTL), with no `I18nManager.isRTL` and no `right`. Also switched the new assign chip to `paddingStart/paddingEnd` so it mirrors in Arabic.

PR #595, opened 2026-08-31
fix(mobile): raise bottom sheets above raised cards, fold the plan add into the day heading

dmadan86:
Two reported UI bugs, both an affordance sitting in the wrong place in the stack.

### The Activity date filter opened behind the feed

Tapping "filter by date" on Activity showed the sheet *underneath* the activity cards.

Being last in the tree is enough on iOS, but not on Android: there a raised view paints above every unraised sibling whatever the tree order. This app's cards are raised — `shadow.soft` is elevation 4, `shadow.lifted` is 10, the pill tab bar is 6 — and `SheetOverlay` had no elevation of its own, so the feed painted straight through the backdrop.

`SheetOverlay` now sits at `elevation: 24` with a matching `zIndex: 100` so the web build orders it the same way. Because it is the shared sheet primitive, this fixes the same latent bug in the currency and group pickers on the capture and add-expense screens.

### The trip plan repeated "+ Add" once per day

Every day section carried a standing `+ Add` row beneath its items, so a screen of eight days wore eight identical buttons — the affordance repeating far more loudly than the plan it belonged to.

The day heading was already the thing being pointed at when somebody wanted to add to that day, so it now does the job: tapping the heading opens the same inline field the row used to. A day with nothing planned, nothing spent and no open field renders no card at all, rather than the empty sliver of surface the add row used to fill.

### Checks

- `tsc --noEmit` clean (the pre-existing `fileURLToPath(URL)` errors in the two vitest configs are untouched and unrelated)
- eslint clean, prettier applied
- full mobile suite: 694 tests / 55 files pass

Not device-tested — the elevation fix in particular is worth a look on a real Android build.

## Summary by CodeRabbit

* **New Features**
  * Daily plan headings are now tappable controls for adding items.
  * Empty day cards are hidden unless they contain planned items, expenses, or an active add form.

* **Bug Fixes**
  * Fixed expense sheets appearing behind cards and navigation elements on Android and web.

### DIFF
--- a/apps/mobile/src/app/group/[id]/plan.tsx
+++ b/apps/mobile/src/app/group/[id]/plan.tsx
@@ -777,9 +777,19 @@ export default function PlanScreen() {
               }}
               accessibilityRole="button"
               accessibilityLabel={`${t.add} — ${dayLabel(day.day, locale)}`}
-              style={({ pressed }) => ({ opacity: pressed ? 0.6 : 1 })}
+              accessibilityHint={t.addPlanHint}
+              // A heading is a line of text, and a line of text is a shorter
+              // target than a finger. Now that this line is the only way to add
+              // to a day it has to be reachable like a button: a full-height row
+              // to aim at, and slop around it for the miss.
+              hitSlop={8}
+              style={({ pressed }) => ({
+                minHeight: 44,
+                justifyContent: 'center',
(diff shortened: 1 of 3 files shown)


dmadan86:
@coderabbit review

PR #596, opened 2026-08-31
fix(voice): one arbiter for the one microphone, so a second capture opens

dmadan86:
Reported: in speech-to-expense, the first time you speak it works; if you want to speak again, the mic is dead.

### Root cause

Both platforms give the app a **single** recogniser object with a **single** global event stream — `ExpoSpeechRecognitionModule.start/stop/abort` and the `result`/`error`/`end` events are the same recogniser no matter which screen asked. And `abort()` does not end a session synchronously: it queues a teardown (Android posts it to the main looper, iOS runs it in a `Task`) whose `end` arrives some milliseconds later.

Nothing attributed events to a session, so on the second attempt the freshly-keyed panel would either:

- receive the previous session's trailing `end` and close its own capture before it had begun, or
- issue its native `start()` before that queued teardown ran — and the teardown then calls `cancel()`/`destroy()` on whatever the recogniser field holds, which is now the *new* one. Silently: no result, no error, no `end`, a mic stuck on "listening".

Two further holes fed the same symptom:

- `start()` was re-entrant. There are two awaits before `listening` flips (permission, installed-model probe), so a double tap — or a tap racing the auto-start — issued two native starts, the second destroying the first.
- `mounted.current` was set `false` on cleanup but never re-armed, so after a Fast Refresh every later `start()` bailed out after its first await.

PR #529 removed the *needless* abort of an already-idle recogniser, but the case where the panel unmounts while genuinely listening still raced, and nothing attributed events to a session.

### The fix

`apps/mobile/src/lib/speechMic.ts` — a React-free, native-module-free arbiter for the one recogniser (the microphone is injected as a two-call driver, so the sequencing is testable without a device). One owner token per session:

- ownership is dropped **synchronously before** `abort()`, so the `error: aborted` and the `end` that follow belong to nobody and are ignored by whichever panel is mounted by then;
- the next `acquire` **waits for the previous session's `end`** rather than opening on top of its teardown;
- a 1.5s guard timer means a missing `end` costs a pause, never the microphone;
- a bounded "owed end" debt swallows a late teardown ending so it cannot close the session that replaced it;
- ownership is deliberately held across `stop()`, which unlike `abort()` still delivers one last `result` — dropping it there would discard the words spoken before the tap.

`VoiceCapture` now gates all five event handlers on `speechMic.owns(session)`, makes `start()` non-re-entrant (claim, then await, release on every exit path including a refused permission and an unmount during the permission sheet), adds a watchdog for a native start that emits nothing at all, and re-arms `mounted` on the way in. The `listeningRef` mirror from #529 is gone — the arbiter is the synchronous source of truth.

`DictateVoice` had its own module-level `micOwner` lock over the same native recogniser — two independent locks on one device is its own bug class — and now shares the arbiter.

### Lifecycle paths covered

Success → immediate restart; user-stop; error; no-speech/timeout; unmount mid-session; unmount during the permission sheet; rapid double-tap; a claim refused because another surface holds the mic (ignored, never stolen); a teardown that never reports; a teardown that reports late after a new capture opened; a native `start()` that produces no events; Fast Refresh.

### Checks

- 14 new tests in `apps/mobile/test/speechMic.test.ts`; full mobile suite 694 tests / 55 files pass
- `tsc --noEmit` clean, eslint clean
- no new user-facing strings, so no i18n change

**Not device-tested** — the emulator has no live mic, so this is verified by reasoning over the sequencing plus unit tests. The restart path is what to exercise first on real hardware.

### DIFF
--- a/apps/mobile/src/components/DictateVoice.tsx
+++ b/apps/mobile/src/components/DictateVoice.tsx
@@ -15,11 +15,13 @@
  *
  * The native recogniser is a single global object with one event stream: every
  * mounted `DictateVoice` (a review screen shows one per expense row) subscribes
- * to the *same* `result`/`error`/`end` events. So a module-level lock hands the
- * mic to one field at a time, and each instance acts on an event only while it
- * is the one listening — otherwise idle rows cross-write the transcript, and a
- * row unmounting (the list changing after "add more") aborts a capture some
- * other surface just started.
+ * to the *same* `result`/`error`/`end` events — and so does the voice quick-add
+ * panel on the other side of the app. So the mic is handed out by the shared
+ * arbiter in `lib/speechMic`, one holder at a time, and each instance acts on an
+ * event only while it is the holder: otherwise idle rows cross-write the
+ * transcript, a row unmounting (the list changing after "add more") aborts a
+ * capture some other surface just started, and the trailing events of a
+ * torn-down session close the one that replaced it.
  */
 
 import { useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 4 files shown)


dmadan86:
@CodeRabbit review

PR #597, opened 2026-08-31
feat(receipts): keep every bill on the phone first, show it going up, resume it after a restart

dmadan86:
Reported: adding a receipt gives no loading indication at all — you can't tell whether it worked. It should not be lost if the app is closed, it should finish on its own, and it should keep going if you move to another screen.

### What was already there, and what was actually wrong

`apps/mobile/src/lib/receiptQueue.ts` already held a durable park-and-send queue: bytes written to a file under the OS document dir, an entry in AsyncStorage, an `attachmentId` chosen up front so the optimistic tile and the real row share an identity, a single-flight `flushReceiptQueue`, a progress bar, orphan-file cleanup, and permanent-vs-transient failure classification.

**But online adds did not use it.** They ran `useAttachExpenseAttachment`, a react-query mutation living on the screen. That is what produced all three complaints at once: nothing rendered until the row landed, leaving the screen dropped the mutation's callbacks, and an app killed mid-upload lost the photograph outright because it existed only in memory.

Compounding it, the only thing that ever flushed the queue was a `useEffect` inside the receipts gallery — so a parked receipt was never sent unless you happened to be looking at that expense.

### The change

**One path now, the durable one.** Every add parks first (`enqueueReceipt`), online or not, so the bytes are on disk before anything is attempted. `useAttachExpenseAttachment` is deleted.

**Live status** through a `useSyncExternalStore` store (`usePendingReceipts`): `queued | uploading | failed`. `uploading` is deliberately *not* persisted, so an app killed mid-flight comes back as `queued` rather than a stuck spinner. `publish` suppresses no-op notifications so a flush that found nothing doesn't re-render every mounted gallery.

**Resume** — `resumeReceiptUploads()` wired into the three lifecycle moments `sync/provider.tsx` already listens for: signed-in launch, foreground, reconnect. This is what makes resume-after-restart and continue-across-navigation real rather than incidental. A success calls `syncEngine.flush()` so the optimistic tile is replaced by the real row.

**The tile says which state it is in** — spinner sending, cloud glyph waiting, filled red disc failed — with a status line under the strip and an inline retry. Tapping a failed tile opens the reason plus Try again / Remove. No raw error text reaches the UI; `lastError` stays on the entry for diagnosis only.

**Retry backoff** so event-driven flushes don't hammer a failing server.

### Behaviour changes worth reviewing

1. **A permanent refusal is no longer silently deleted.** Cap reached or not-a-party now keeps `permanent: true` with the bytes intact and shows as a failed tile, because the usual fix for the cap is to remove another receipt and re-send this one. Background flushes skip it; `retryPendingReceipts` clears the mark, the backoff and the error. `FlushResult` gained `capReached` so the screen can offer the upgrade instead of an unexplained red tile.

2. **The visibility prompt is gone.** Photographing a bill used to stop and ask "Who can see this receipt?" a beat after you were done. The answer was the group's every time — a receipt for a bill everybody is paying a share of is not a private document — and group-readable is what the R2 group path already authorises by membership. The narrower `'parties'` visibility stays in `commitAdd` and in the storage rules for payment proofs, and a receipt already added that way keeps its Private tag.

### Checks

- 7 new i18n keys in the typed interface and all four locales (en/ta/hi/ar); RTL-safe — status overlay centred, strip uses `Row`
- `apps/mobile/test/receiptQueue.test.ts`: 14 pass (3 pre-existing privacy tests preserved, 11 new). The cold-start test uses `vi.resetModules()` + re-import to prove resume works with no in-memory state
- full mobile suite 694 tests / 55 files pass; `tsc --noEmit` clean; eslint clean

**Not device-tested** — I can't kill and relaunch a real app from here, so resume-on-restart is proven by the cold-start test rather than on hardware. That's the path to exercise first.

### DIFF
--- a/apps/mobile/src/components/ExpenseReceipts.tsx
+++ b/apps/mobile/src/components/ExpenseReceipts.tsx
@@ -13,7 +13,7 @@
  * orphaned; new images are all attachment rows.
  */
 
-import { forwardRef, useCallback, useEffect, useImperativeHandle, useMemo, useState } from 'react';
+import { forwardRef, useEffect, useImperativeHandle, useMemo, useState } from 'react';
 import Ionicons from '@expo/vector-icons/Ionicons';
 import { Image } from 'expo-image';
 import { router } from 'expo-router';
@@ -41,7 +41,6 @@ import { ReceiptAnnotator } from '@/components/ReceiptAnnotator';
 import { ReceiptCropper } from '@/components/ReceiptCropper';
 import {
   useAnnotateExpenseAttachment,
-  useAttachExpenseAttachment,
   useExpenseAttachments,
   useRemoveExpenseAttachment,
   useRemoveExpenseReceipt,
@@ -56,10 +55,12 @@ import {
   discardPendingReceipt,
   enqueueReceipt,
   flushReceiptQueue,
-  isOnline,
-  listPendingReceipts,
   pendingReceiptUri,
-  type PendingReceipt,
+  retryPendingReceipts,
+  usePendingReceipts,
+  type FlushResult,
+  type PendingReceiptStatus,
+  type PendingReceiptView,
 } from '@/lib/receiptQueue';
 import { SyncStatus, useSync } from '@/sync';
 import { fill, useStrings } from '@/i18n';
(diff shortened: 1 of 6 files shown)


dmadan86:
@CodeRabbit review

PR #599, opened 2026-08-31
chore(mobile): take the SDK 57 patches, and let the debug build sit beside the release one

dmadan86:
Two things: the SDK 57 patch upgrades Expo was nagging about, and a fix for the debug and release APKs replacing each other on a phone.

## The upgrades

`expo install --check` listed 31 packages. **23 are genuine patch upgrades inside SDK 57** and are taken here — `expo 57.0.12 → 57.0.18`, `expo-router → 57.0.17`, `expo-updates → 57.0.19`, `react-native 0.86.2 → 0.86.3`, and the rest of the `expo-*` set.

**8 were downgrades** — Expo's SDK-pinned expectations against versions this repo runs ahead of on purpose — and are left alone:

| package | installed | Expo wanted |
| --- | --- | --- |
| `@react-native-async-storage/async-storage` | 3.1.1 | 2.2.0 |
| `@sentry/react-native` | 8.22.0 | ~7.11.0 |
| `@shopify/flash-list` | 2.3.2 | 2.0.2 |
| `react` / `react-dom` | 19.2.8 | 19.2.3 |
| `react-native-reanimated` | 4.5.3 | 4.5.1 |
| `react-native-safe-area-context` | 5.8.1 | ~5.7.0 |
| `react-native-screens` | 4.27.0 | ~4.26.0 |

async-storage 3 was a deliberate choice in #178; the rest are the same "stay current" call. `expo install --check` will keep listing these nine — that is expected, not drift.

RN 0.86.2 → 0.86.3 is a patch inside 0.86 and unrelated to the 0.87 attempt that had to be reverted in #200.

One real breakage found: `packages/ui` still pinned `react-native 0.86.2` in its devDependencies, so bumping the app alone put two copies of the RN types in the graph and `tsc` failed with `Type 'TextStyle' is not assignable to type 'StyleProp<TextStyle>'` — the same type, from two paths. Both now say 0.86.3.

## Side-by-side debug and release

Only one APK could be installed at a time: Android identifies an app by its application id, both builds claimed `app.waves.mobile`, so installing the debug build replaced the release build that was there to be compared against.

The debug variant now takes `.debug` on its application id, `-debug` on its version name, and `Waves dev` as its launcher label. Verified on the built APKs:

```
debug    app.waves.mobile.debug   "Waves dev"   0.1.0-debug   327 MB
release  app.waves.mobile         "Waves"       0.1.0         201 MB
```

### Why the Firebase task is skipped for debug

The Google Services gradle plugin fails a build outright when `google-services.json` has no client for the id being built:

```
No matching client found for package name 'app.waves.mobile.debug'
```

`applicationId` cannot be set per build type (only suffixed), so pointing the debug variant at the already-registered `app.baaki.mobile` client was not an option either — that needs product flavors. So the debug variant skips `processDebugGoogleServices` and builds with no Firebase config, which is the **same no-push mode `app.config.ts` already supports** for anybody building this repo without a Firebase account, and which `lib/push.ts` already degrades into rather than throwing.

**Push therefore does not work in the debug build.** To give it push: register an Android app for `app.waves.mobile.debug` in the Firebase console, re-download `google-services.json` (it comes back carrying both clients), and delete the `whenTaskAdded` block. The plugin header says so.

### Why a config plugin

`apps/mobile/android/` is generated and untracked, so an edit to `build.gradle` would vanish at the next `expo prebuild`. `plugins/withSideBySideDebug.js` applies the gradle change and writes the debug source set's `strings.xml` (a resource in a build type's source set replaces `main`'s for that variant only, which is what keeps the release build named `Waves`).

## Known caveat, not fixed here

Both apps declare the `waves://` and `baaki://` schemes, so with both installed an OAuth redirect or an invite link raises an app chooser. Narrowing the debug variant's intent filters would break Google sign-in for it, so it is left as is.

## Checks

Workspace `typecheck` (9 projects) and `lint` green. Mobile tests 701 passed. Both APKs built locally with `gradlew` and their ids read back with `aapt2 dump badging`.

`packages/db` tests fail in this environment — all 55 files with `client` undefined, i.e. no local Postgres on :54330. Unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added support for installing Android debug builds alongside release builds, with distinct app identifiers, version labels, and launcher names.
  * Debug builds can now work without matching Firebase client configuration.

* **Bug Fixes**
  * Updated mobile framework and Expo components to newer patch versions for improved stability and compatibility.


dmadan86:
@CodeRabbit review

PR #604, opened 2026-09-01
feat(expense): one bill, several payers — and a form that looks like the screen it edits

dmadan86:
## Why

The ledger has always been multi-payer. `expense_payers` is a table, a balance is `Σpaid − Σowed` over every payer row, the mirror replays a payer map, and both edge functions reject a write whose rows do not sum to the total (`PAYER_MISMATCH`). Only the clients collapsed it — one `payer: MemberId | null`, written out as `{ [payer]: amount }`.

So "she got the taxi, I got the tickets" had to be entered as two expenses: two rows in the feed, two things to edit, two things to delete, for one dinner.

**No migration and no edge deploy.** This is the client catching up to a schema that was already there.

Grew out of a UI pass on the expense screen (edit was hard to find, view and edit felt like different apps, the amount ate the first screenful), so those fixes ride along.

## Core — `@waves/core`

New `split/payers.ts`, pure and integer-only:

| | |
|---|---|
| `splitPaidEqually` | Reuses `splitEqually`, so the payer side rounds by the same ADR-009 rule as the share side. Two rounding rules on one row would fail its own trigger. |
| `validatePayers` | Mirrors the server's check — a mismatch becomes a sentence under the button, not a `PAYER_MISMATCH` off a queue minutes later. |
| `rebalancePayers` | "Typed figures are facts, the rest is arithmetic." Locked payers survive verbatim; unlocked ones share what is left. |
| `serialisePayers` | Wire form, zero rows dropped. |

`rebalancePayers` makes two deliberate refusals: nothing is ever driven negative, and locked figures are never rescaled to fit. Both would make a row add up while describing something that did not happen.

New `money/input.ts` — the text↔minor parsing lifted out of `AmountField`, because the per-payer fields parse the same keystrokes. Two copies of "how many decimal places does this currency have" is how ₹10.5 becomes 105 paise on one screen and 1050 on another. `AmountField` now imports it.

## Form

One payer stays **exactly one tap** — the avatar row, no figures. An explicit **"Several people paid"** switch turns it into a multi-select with per-payer amount fields, a "Split evenly" reset, and a running "₹200 left to assign".

The mode is explicit because the gestures are opposite: with one payer a tap must *replace* (the "actually she paid" correction was one tap before this and stays one tap), with several it must *add*. `accessibilityRole` follows the mode — radio vs checkbox — so a screen reader is never told the opposite of what will happen.

Rebalancing runs as a render-phase adjustment, the pattern this file already uses for seeding, so a receipt scan filling in ₹1,240 re-splits the paying in the same render — no frame where the payers and the total disagree.

Two bugs fixed on the way:

- **Editing flattened a multi-payer bill.** It hydrated `payers[0]` and saved that back, silently rewriting who had put money in.
- **Who-paid was hidden entirely when editing** (`{!expenseId ? … : null}`), so the correction people most often return to make had no control anywhere on the screen while the summary kept saying "You paid".

## Expense screen

**"Who owes what" was lying on a multi-payer bill.** Every row was forced owe-red, on the reasoning that every share is money owed — true while one person could pay. Somebody putting in ₹5,000 of a ₹10,000 bill and owing ₹2,000 read as *owing* ₹2,000 when they are owed ₹3,000.

Rows now show **net for the bill (paid − share)**, sign-derived. One rule, no special case: a non-payer's net is exactly minus their share, so ordinary single-payer bills look unchanged. Rows are built from the union of shares and payers, so a contributor who owes nothing is no longer visible in "Paid by" and then unaccounted for below it.

Also on this screen:

- **Edit** was the first item of a three-dot menu; it is a pencil in the header now. Delete stays a tap deeper.
- **Split type carries a glyph**, shared with the form's chips, so the chip you tapped is the mark you see afterwards.

## Form and screen now share a header

Same brand panel, same category badge in the same corner, same amount — editable on one side, not on the other. Tapping Edit changes what you can do, not where anything is.

That also settles what the amount costs: it and its currency share one line inside the header instead of a 44pt number owning a third of the first screenful, with the running total already in the pinned save bar.

## Smaller things in here

- The expense screen's **receipt gallery** replaces the one-line thumbnail on the edit form — existing expenses only, since attachments commit against a row a new expense does not have yet.
- The **split section is tightened** so who-paid and who-splits fit one screen.
- The **"nothing is overwritten" banner** is gone from the edit form.
- The **edit history** gains a coloured "Your share" line, driven by the activity feed's own `myStake` so the two cannot drift. Amount changes stay neutral ink — a total belongs to nobody, and colouring it by sign would have the screen reader announce "you are owed ₹10,000" about a dinner.
- New strings in **en / ta / hi / ar**.

## Tests

**55 new in `@waves/core`**, three layers because the failure modes differ:

**Properties (fast-check)** — sums to the total for every input; never spreads two payers more than one minor unit; deterministic across devices; never negative whatever the locked figures; only ever returns the selected payers; idempotent.

**Gestures** — type one figure and the other moves; two typed and the third absorbs; raise the total after the fact; overshoot is kept and reported rather than trimmed; drop a payer; collapse to one; odd amounts; a currency with no minor unit.

**Scenarios** — whole expenses through `computeShares` + `computeNetBalances`, and `payers.offline.test.ts` runs the ADR-005 path (mirror + queue overlay, the pipeline behind the group hero number): a multi-payer bill queued with no network nets correctly before it reaches a server, is one row not one-per-payer, survives the wire round trip, re-versions on edit without duplicating, and a payer who is not in the split shows up as a creditor.

The assertion worth pointing at: `validatePayers` is checked against an **independently written** copy of the server's rule, so client and server cannot drift.

## Gates

`pnpm typecheck`, `pnpm lint`, and every suite except `@waves/db` — **1573 tests green**.

`packages/db` has 27 pre-existing failures (RLS / anon-grant state on the local Postgres). I stashed this branch and reproduced them unchanged on `main`; they are not from this work.

## Not in here

- **`apps/web/src/components/ExpenseForm.tsx` still writes a single payer.** The core functions are ready for it; I did not expand a large mobile change into a second codebase uninvited.
- **Not device-tested.**

## Summary by CodeRabbit

- **New Features**
  - Added support for expenses paid by one or multiple people, with editable payer amounts and automatic balancing.
  - Improved expense entry with a redesigned header, currency controls, receipt galleries, and clearer split options.
  - Enhanced expense details with payer/share breakdowns, net balances, icons, and easier access to editing.
  - Added clearer history updates showing how changes affect your share.
  - Added localized payer and audit-history text in English, Tamil, Hindi, and Arabic.
- **Bug Fixes**
  - Improved receipt display by hiding empty gallery headings.
  - Improved money entry accuracy across currencies.

### DIFF
--- a/apps/mobile/src/app/capture.tsx
+++ b/apps/mobile/src/app/capture.tsx
@@ -17,6 +17,7 @@ import {
   type ExpenseLocation,
   type HeuristicReceipt,
   type PaymentMethod,
+  byNewest,
 } from '@waves/core';
 import {
   Button,
@@ -861,7 +862,7 @@ function GroupPicker({
   // Newest-created first, standing in for "recently used": the group model
   // carries no updated_at / last-activity field, so creation order is the only
   // honest recency signal available without a heavier query.
-  const byRecent = [...rest].sort((a, b) => (b.created_at ?? '').localeCompare(a.created_at ?? ''));
+  const byRecent = [...rest].sort(byNewest((row) => row.created_at));
 
   // Only split "recently used" out from "all" once there are enough groups that
   // a shortcut to the top few actually saves scrolling.
--- a/apps/mobile/src/app/group/[id]/add-expense.tsx
+++ b/apps/mobile/src/app/group/[id]/add-expense.tsx
@@ -6,6 +6,7 @@ import { router, useLocalSearchParams } from 'expo-router';
 import { StatusBar } from 'expo-status-bar';
 import {
   ActivityIndicator,
+  Alert,
   KeyboardAvoidingView,
   Platform,
   Pressable,
@@ -24,11 +25,9 @@ import {
   guessCategory,
   money,
   MutationKind,
-  parseMinorInput,
   PayerProblemCode,
   rebalancePayers,
   ridersSplit,
-  sanitiseMinorInput,
   serialisePayers,
   splitByUnits,
   treatSplit,
@@ -81,11 +80,20 @@ import { receiptCapStatus, receiptTapAction } from '@/lib/receiptCapGate';
 import { StorageCapError } from '@/lib/storage';
 import { useAssignCapture, useGroup } from '@/data/hooks';
 import { displayName, groupLabel, isGhost } from '@/data/types';
-import { plural, useStrings } from '@/i18n';
+import { fill, plural, useStrings } from '@/i18n';
 import { useAuth } from '@/lib/auth';
 import { useGuestGuard } from '@/lib/guestGuard';
 import { handoverKey } from '@/lib/handover';
 import { resolveDraftCurrency, resolveDraftFx } from '@/lib/expenseDraft';
+import {
+  expenseDateFor,
+  planCollapseToOne,
+  planEvenly,
+  planToggle,
+  planTypedAmount,
+  todayIso,
+  type PayerPlan,
+} from '@/lib/expenseForm';
 import { captureReceipt, pickReceiptImage, type PickedImage } from '@/lib/image';
 import { recogniseReceipt } from '@/lib/ocr';
 import { matchMemberNames, stripMemberNames } from '@/lib/voiceExpense';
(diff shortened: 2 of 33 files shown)


dmadan86:
@CodeRabbit review

dmadan86:
@CodeRabbit review

PR #644, opened 2026-09-04
fix(email): surface unknown provider configuration

dmadan86:
## Summary
- Makes unknown EMAIL_PROVIDER values surface as misconfiguration instead of silently treating email as off
- Adds focused edge tests for the email provider seam across Resend and SendGrid
- Covers traveller settlement, rider notification, and financer campaign send paths, including SendGrid custom_args and transient/permanent failure classification

## Validation
- pnpm test:edge -- _shared/email.test.ts
- pnpm exec prettier --check supabase/functions/_shared/email.ts supabase/functions/_shared/email.test.ts
- pnpm lint
- Attempted pnpm typecheck; blocked locally by existing mobile route type errors unrelated to this email seam change

### DIFF
--- a/apps/mobile/src/components/AuthFlow.tsx
+++ b/apps/mobile/src/components/AuthFlow.tsx
@@ -430,7 +430,7 @@ export function AuthFlow({ flow }: { flow: AuthFlowKind }) {
               busy={busy}
               onGoogle={() => void run(withGoogle)}
               onApple={() => void run(withApple)}
-              onPhone={() => router.push('/phone')}
+              onPhone={isSignup ? undefined : () => router.push('/phone')}
               t={t}
             />
             {/* ADR-006 addendum: the guest way in belongs to the sign-up page —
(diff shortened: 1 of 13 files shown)


dmadan86:
Superseded by #641, which now carries the substance of this PR.

**Taken:** the new `_shared/email.test.ts`. That was the real find — `email.ts` had no coverage of its own, on the reasoning that the logic lives in `@waves/core`. True of the old file; untrue the moment it had to choose a provider. Ported onto the current API in 97fefc0 and passing.

**Also already in #641:** surfacing an unknown `EMAIL_PROVIDER` instead of reading it as "email is off". #641 goes a step further and carries a *reason* rather than throwing, so `notify-fanout` can tell a misconfiguration (report it, claim nothing) from a deployment that simply has email switched off (silent, not a fault).

**Not taken, and the reason to close rather than merge:** this branch predates #639 and #640, so it re-adds `otp-send/*` and the mobile auth files wholesale. Merging it would revert the `v1,` hook-secret strip — without which every genuine request is answered 401 — along with the refused-connection catch and the config-check-before-quota fix from #640. Same stale-base pattern #640 had.

Closing in favour of #641.

PR #669, opened 2026-09-06
fix(sync): a group does not need a name

dmadan86:
## What broke

Create a group without typing a name and it never arrives. The phone opens **"Group not found"** for the group it has just made, the dashboard goes back to **"No groups yet"**, and the only clue is a red alert glyph in the header. Reproduced from a released build; confirmed against prod, where the account had 0 groups after a create.

`new-group.tsx` treats a name as optional and says so in as many words:

```ts
// Blank is fine — the group gets labelled by who is in it instead.
name: name.trim() || null,
```

`waves_create_group` accepts a null name too. The `sync` function did not:

```ts
p_name: requireString(mutation.payload.name, 'name'),   // → VALIDATION_FAILED
```

So the mutation was refused before it reached the database.

The failure is loud because `group.create` has **no optimistic mirror row** — the group exists on the phone only once the server's change comes back. Queued, refused, never materialised. Nothing on the screen names the cause.

## What this changes

`p_name` goes through a new `optionalString`, which also collapses a blank or whitespace-only name to `null` rather than passing `''` on: an empty string in `groups.name` is a name that renders as nothing everywhere, instead of falling back to the members.

Every other `requireString` in the file was checked — they guard ids, a day, a title, an amount, and a ghost member's name, all genuinely required. Only this one was wrong.

## Verified

`supabase/functions/sync/groupCreate.test.ts` drives `SyncSession` with a stubbed client: `null` and `'   '` both reach the RPC as `p_name: null`, and `'  Goa trip  '` still arrives trimmed. `pnpm test:edge` — 7 files, 112 tests, green.

Branched off `main`, independent of #667 and #668.

## Summary by CodeRabbit

- **New Features**
  - Group creation now allows names to be omitted.
  - Blank or whitespace-only group names are stored as empty values.
  - Group names are automatically trimmed when creating or updating groups.

- **Bug Fixes**
  - Updating a group with a blank name now correctly clears the existing name.
  - Unrelated group update fields remain unchanged when the name is modified.

### DIFF
--- a/supabase/functions/sync/groupCreate.test.ts
+++ b/supabase/functions/sync/groupCreate.test.ts
@@ -1,0 +1,112 @@
+/**
+ * A group does not need a name.
+ *
+ * `new-group.tsx` says so in as many words — "Blank is fine — the group gets
+ * labelled by who is in it instead" — and it sends `name: null` when nobody
+ * typed one. `waves_create_group` accepts that. This function did not: it ran
+ * the name through `requireString`, so a nameless group was rejected with
+ * `VALIDATION_FAILED: name is required` before it ever reached the database.
+ *
+ * The cost was worse than a failed save, because `group.create` has no
+ * optimistic mirror row — the group only exists on the phone once the server's
+ * change comes back. So the create was queued, refused, and never materialised:
+ * the phone opened "Group not found" for the group it had just made, the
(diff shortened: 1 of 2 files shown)


dmadan86:
@coderabbitai review

PR #670, opened 2026-09-06
fix(feedback): make "sent" the whole screen, not a notice above an empty one

dmadan86:
## Before

The form vanishes on send and leaves a small card at the top of the page it just vacated — a tick and a Close button stranded above a whole screen of nothing. It reads less like an acknowledgement than like a screen that half-loaded.

The tick was also `theme.color.positive`, which in dark mode is `#60A5FA` — the *money* blue that means "you are owed". A success mark borrowing the balance colour on a screen with no money in it says something it does not mean.

## After

The confirmation is the only thing on screen at that moment, so it occupies it: centred, one mark in a `brandSoft` disc, the thanks, and a line saying what actually happens to what they wrote.

- **`brand`, not `positive`** — see above.
- **Two actions, not one.** Done, and *Send another*, which returns an empty form (message, rating and kind all cleared — a second thought is a new message). The thought that arrives right after sending used to cost a trip back out through settings and in again.
- **A small spring entrance on the tick**, so it reads as a reply to the tap rather than something that was always there. Skipped entirely under `useReducedMotion`.
- Two new strings, `feedbackThanksBody` and `feedbackAnother`, in **en / ta / hi / ar**.

The compose form itself is untouched.

`pnpm typecheck` and `pnpm lint` clean in `apps/mobile` (3 pre-existing `phone.tsx` warnings); prettier clean. Not device-tested. Branched off `main`, independent of #667/#668/#669.

### DIFF
--- a/apps/mobile/src/app/settings/feedback.tsx
+++ b/apps/mobile/src/app/settings/feedback.tsx
@@ -1,8 +1,16 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import Ionicons from '@expo/vector-icons/Ionicons';
 import Constants from 'expo-constants';
 import { router } from 'expo-router';
-import { ActivityIndicator, Platform, Pressable, ScrollView, TextInput, View } from 'react-native';
+import {
+  ActivityIndicator,
+  Animated,
+  Platform,
+  Pressable,
+  ScrollView,
+  TextInput,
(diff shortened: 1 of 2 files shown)


dmadan86:
@coderabbitai review

dmadan86:
@coderabbitai full review

PR #673, opened 2026-09-06
fix(sync): validate budget payload clears

dmadan86:
## Summary
- validate nullable budget/rate payload fields at the sync edge before dispatching RPCs
- keep explicit null as the clear operation while rejecting missing or non-text values
- add user/rider/traveller/financer-style coverage for trip budget and FX-rate payloads

## Tests
- pnpm test:edge


dmadan86:
@coderabbitai review

PR #674, opened 2026-09-06
fix(r2): require party for restricted reads

dmadan86:
## Summary
- require subject group membership and explicit party status before signing restricted R2 reads
- keep existing party-visible row/path check as the final proof for exact object visibility
- add coverage for user/rider/traveller/financer-style restricted attachment/proof reads

## Tests
- pnpm test:edge


dmadan86:
@coderabbitai review

PR #676, opened 2026-09-06
fix(web): stop showing the database's own words to whoever opened the link

dmadan86:
Follow-up to the balance-card fix (#675) — a sweep for other places the app shows a user its own internals.

## What the sweep found

| Surface | Verdict |
| --- | --- |
| **`apps/web` raw errors** | **Broken — fixed here.** 18 sites rendered `caught.message` verbatim. |
| Group balance cross-check card | Fixed in #675. |
| `apps/mobile` error surfaces | Clean. Everything routes through `friendlyError` since #256; the three remaining `.message` reads only *test* for `NUDGE_RATE_LIMIT` and render translated strings. |
| `SyncBanner` refused / stuck cards | Correct as they are — a change of yours did not save, with Try again / Let it go beside it, and "still saved on this phone" so nobody thinks their entry is lost. |
| `apps/admin` raw errors | Left alone on purpose. Staff console; the Postgres sentence is the useful one there. |

## The web problem

Every `catch` ended in `setError(caught instanceof Error ? caught.message : String(caught))`. What that puts in front of somebody who installed nothing and followed a link:

```
new row violates row-level security policy for table "expenses"
Could not find the function public.waves_create_group in the schema cache
JSON object requested, multiple (or no) rows returned
JWT expired
```

Untranslated, unactionable, and on the join page wedged between "This link is broken" and the sentence that already explains it. Worse here than on mobile: this is the app for a first-time guest, and a page that talks about schema caches is a tab they close.

## The fix

`apps/web/src/lib/errors.ts` — the web `friendlyError`, mirroring `apps/mobile/src/lib/errors.ts`. Separate files on purpose: that one imports from the React Native module graph, and what is shared is the doctrine, not the code.

- a rate limit is not a rejected credential — "too many tries in a row", not "could not sign in";
- a dropped connection gets named, but in our words, never the transport's;
- everything else goes to Sentry tagged `where`, and the page gets a plain fallback.

Five new strings (`couldNotLoad`, `couldNotSave`, `couldNotSignIn`, `offline`, `tooMany`) in all four languages. The raw line on the broken-invite card is deleted.

## Gates

`typecheck` 0 · `lint` 0 · web tests **33 passed** (4 files, +6 covering each raw shape seen in the wild)

### DIFF
--- a/apps/web/src/app/add/page.tsx
+++ b/apps/web/src/app/add/page.tsx
@@ -23,6 +23,7 @@ import { Section } from '@/components/Shell';
 import { SkeletonRows } from '@/components/Skeleton';
 import { plural } from '@/i18n';
 import { useStrings } from '@/i18n-context';
+import { friendlyError } from '@/lib/errors';
 
 export default function AddPage() {
   return (
@@ -59,7 +60,13 @@ function Picker({ profileId }: { profileId: string }) {
         setGroups(g);
         setMembersByGroup(m);
       } catch (caught) {
-        if (active) setError(caught instanceof Error ? caught.message : String(caught));
+        if (active)
+          setError(
+            friendlyError(caught, 'web.add.load', {
+              fallback: t.errors.couldNotLoad,
+              offline: t.errors.offline,
+            }),
+          );
       } finally {
         if (active && !forwarding) setLoading(false);
       }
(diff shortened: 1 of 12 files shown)


dmadan86:
@coderabbitai review

PR #682, opened 2026-09-07
feat(email): a sign-in alert, a Monday digest, and one sending domain

dmadan86:
Three mails the app promised and had nothing behind.

## `new_device_login` — a sign-in worth telling somebody about

Written by `waves_register_device` the first time a device id is seen on an account, and **again when a signed-out device comes back** — that second case is somebody who revoked a device deciding it wasn't theirs.

It's the only *security* notice in the notification set, and it's treated as one in three places:

- it **ignores `notification_prefs.email`** — turning off ledger mail is not a request to stop hearing that your account was opened somewhere;
- it **never waits on push** — unlike `group_added`, which mails only as a fallback. The device is the thing in question, and a stolen phone receives its own alert;
- the mail carries **no unsubscribe line and no `List-Unsubscribe` header** (`SECURITY_TEMPLATES` in `@waves/core`). That's not dodging bulk-mail rules: RFC 8058 governs bulk mail, and a sign-in alert is transactional — the same class as a password reset, which no sender offers an opt-out on.

A hard suppression (bounce, complaint, unsubscribe) still wins — that's the mailbox refusing rather than the person choosing. Guests are skipped: an anonymous account has no address to warn and no password to change.

## `digest_weekly` — Monday morning, seven days

A new pg_cron job at 03:30 UTC on Mondays (09:00 IST) writes one row per account **with activity in the last seven days**, and nothing for a quiet week. An empty digest is how a sender teaches people to write a filter rule.

**It is opt-in.** Settings → Notifications has said "Weekly email digest … Off by default" since before anything produced one, and `DEFAULT_NOTIFICATION_PREFS.weeklyEmail` is `false`, so the producer gates on that flag. A digest that shipped to everybody would have made that screen a lie on the day it started working.

The net figure is in the account's own currency only. Summing currencies would be a lie; picking a "dominant" one would be a lie that changes week to week.

`digest_daily` keeps its template and its lack of a producer, so either cadence can be turned on without the other.

## One sending domain, and auth mail through it

`[auth.email.smtp]` points GoTrue at `smtp.resend.com:465` as `Waves <hello@wavs.co.in>`. Without it GoTrue uses Supabase's built-in sender, which is rate-limited and documented as for testing — fine until two people sign up in the same minute, and then it's a login door that silently doesn't open.

`EMAIL_FROM` defaults to `hello@wavs.co.in`, superseding `mail.dmadan.com`. Two sending domains means two reputations to keep, and the one nobody watches is the one that starts landing in spam.

## Tests

- `packages/db/test/device-alert-and-digest.test.ts` — 11 cases: first sighting writes one alert, a heartbeat writes none, a revoked device returning writes another; the alert survives `email: false` and is still suppressed by a bounce; the digest is per-account, idempotent inside a week, silent for a quiet week, and silent for an account that never opted in.
- `packages/core/test/email.test.ts` — the security template carries no unsubscribe in header, HTML or text, explains itself by the sign-in rather than by a group, and ordinary mail keeps its unsubscribe intact.
- Full suites: core 914, db 729. Migration applied and exercised against a local Postgres.

## Deploy notes

Needs, in order: `RESEND_API_KEY` + `EMAIL_FROM` as edge secrets; `RESEND_SMTP_PASSWORD` in `supabase/.env` then `config push` (`secrets set` does **not** reach GoTrue); `migrate deploy`; `edge:deploy`. The cron job schedules itself from the migration, guarded on pg_cron being present so a local stack skips it.

`EMAIL_UNSUBSCRIBE_SECRET` is also unset on the project — the fanout refuses to claim a row without it.


dmadan86:
Pushed a follow-up after seeing what the mail actually looks like: the sign-in mail Supabase sends says **"Your sign-in link"** and contains a link, while the app asks for the code on a six-box screen. The one thing that screen needs is the one thing the mail doesn't have.

`sendEmailOtp` in `lib/auth.tsx` has a comment that has been wrong since it was written — "Supabase mails a six-digit code (not a magic link) as long as the email template carries `{{ .Token }}`" — and nothing ever carried it.

The five GoTrue templates now live in `supabase/templates/` and go up with `config push`:

| template | subject | why it's here |
|---|---|---|
| `magic_link` | Your Waves sign-in code | email sign-in for an existing account |
| `confirmation` | Confirm your email for Waves | `shouldCreateUser: true` on the sign-up door |
| `email_change` | Confirm your new Waves address | Settings → Account, adding or changing an address |
| `recovery` | Reset your Waves password | no caller today |
| `reauthentication` | Confirm it is you | no caller today |

Every one prints `{{ .Token }}` and **none of them links**. A mail carrying both would teach people to click the link in a message asking for a code, which is the shape of every phishing mail they'll ever get; each also says in the footer that Waves will never ask for the code. The last two are included because GoTrue sends its own link-shaped default the moment a caller appears, and that failure would be silent.

One deploy hazard now written above the block in `config.toml`: `config push` sends the file as it stands, so pushing before `RESEND_SMTP_PASSWORD` is in `supabase/.env` enables SMTP **with an empty password** — which sends nothing at all, worse than the built-in sender it replaces. Key first, then push.

PR #715, opened 2026-09-08
Apply a tag the moment it is made, and stop the value vanishing

dmadan86:
Two regressions from #713, both in the "more details" list on add-expense.

### A new tag was lost on the way back

Tapping "＋ New tag" in the category sheet opens the tag editor. `useUpsertTag` returns the id it just created, but `TagEditorSheet` discarded it (`onSuccess: onClose`), so the editor closed onto a row still showing the *old* category — the tag you had just invented was nowhere on screen, and only findable by opening the sheet again.

This was survivable before #713 because the chip lane stayed visible and the new chip simply appeared in it. Folding that lane into a single row took the affordance away without replacing it.

`TagEditorSheet` now takes an optional `onCreated(tagId, meta)` and add-expense applies what you made. The snapshot is built from the fields as saved rather than read back from the catalog — the upsert is queued locally and the catalog mirrors it, so reading it there would be a race for a value the form already holds exactly. Never fired for an edit, and optional, so the tags manager and the capture screen are untouched.

### `SettingRow` drew no value at all when the label ran long

The value's side was `flex: 1`, which in React Native means `flexBasis: 0`. Once the label's own text outran the row, free space went negative, `flexGrow` stopped applying, and the value settled at **zero width** — the row rendered the question and the chevron with nothing between them. That is worse than a truncated answer: the setting looks unset.

Both halves now carry a content basis (`flexBasis: 'auto'`) with `flexShrink: 1` and `minWidth: 0`, so the shrink is shared and each keeps a readable fraction.

Reproduces with a long Tamil label, and in any language at large font scale. The stale comment claiming the two sides "clip rather than wrap" has been corrected — clipping was never guaranteed.

### Checks

`pnpm --filter @waves/mobile typecheck` clean, `pnpm lint` clean (3 pre-existing warnings in `phone.tsx`), 957 mobile tests pass, Prettier clean across `apps/`, `packages/`, `supabase/`.

Nothing that gets saved to an expense changed.

### DIFF
--- a/apps/mobile/src/app/group/[id]/add-expense.tsx
+++ b/apps/mobile/src/app/group/[id]/add-expense.tsx
@@ -2506,8 +2506,21 @@ export default function AddExpenseScreen() {
         />
       ) : null}
 
-      {/* Make a tag on the spot, from the category sheet's "＋ New tag" row. */}
-      <TagEditorSheet open={editingTag} onClose={() => setEditingTag(false)} />
+      {/* Make a tag on the spot, from the category sheet's "＋ New tag" row —
+          and then wear it. Somebody who breaks off from tagging an expense to
+          invent a tag wanted that tag on this expense; the editor used to close
+          onto a row still showing the old category, with the thing they had
+          just made nowhere on screen and only findable by opening the sheet
+          again. Selecting it here is the rest of that one gesture. */}
+      <TagEditorSheet
+        open={editingTag}
+        onClose={() => setEditingTag(false)}
+        onCreated={(tagId, meta) => {
+          setCategory(tagId);
+          setCategoryMeta(meta);
+          setCategoryChosen(true);
+        }}
+      />
     </Screen>
   );
 }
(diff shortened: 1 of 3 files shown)


dmadan86:
Two corrections to this PR's description, posted after the merge.

**The Tamil / large-font-scale claim was not observed.** The body says the zero-width value "reproduces with a long Tamil label, and in any language at large font scale." That was derived from React Native's flex semantics — `flex: 1` implying `flexBasis: 0`, so negative free space leaves nothing for `flexGrow` to distribute — not from running the screen. I did not exercise it on a device or emulator, in any locale or at any font scale. The fix is reasoned, typechecked and unit-tested at the suite level, but its visible behaviour is unverified. Treat it as such until someone puts it on a phone with `ta` and a large accessibility text size.

**CodeRabbit did not review this PR.** It was rate-limited ("next included review available in 27 minutes") and posted only that notice; the green CodeRabbit check is that notice, not an approval. This PR has zero reviews, as do #711, #712, #713 and #714 — only #710 has real CodeRabbit coverage. A green CodeRabbit check on this repo currently cannot be distinguished from a passing review, which is worth knowing before treating one as sign-off.

PR #717, opened 2026-09-08
Rebuild the marketing site around the ledger it is selling

dmadan86:
The site was competent and generic: a purple-nebula gradient, glass cards, a logo marquee containing no logos, a stats bar with two numbers that are not measurements, six identical alternating feature rows, and 9,900 pixels of scroll in which every section had the same shape. It was also, from the hero card down, an Indian product with an English page — a Goa trip in rupees, UPI listed first among payment rails, "₹0 forever" as the price of free, and `priceCurrency: "INR"` in the structured data.

The subject is a shared ledger, so the page is now built from a ledger's own materials: a true-neutral ground instead of a tinted night sky, hairlines instead of glass, radii small enough to read as ruled boxes, and every figure on the page set in a monospace with tabular lining numerals.

## Design system

**Light first, with a real dark theme.** Every surface and ink colour is a `--w-*` custom property redefined in three places — bare `:root`, `prefers-color-scheme: dark` guarded by `:not(.theme-light)`, and `.theme-dark` — so the OS decides by default and the header's three-state toggle wins over it in either direction. An inline script resolves the stored choice before first paint. No component names a colour literal; if one did, a theme would already be broken.

**Three typefaces, all SIL OFL 1.1**, self-hosted from `src/fonts` with the licence text beside each file: **Overused Grotesk** for prose, **IBM Plex Mono** for data, **Departure Mono** for the wordmark. The last two are Latin-only, so neither is ever pinned to a translated string; Tamil, Devanagari and Arabic keep their own Noto faces, and only the script the page is written in is loaded.

Colour is spent in two places only: the violet the app already uses, and the semantic money pair — which never travels without a sign and a word, so it survives colour blindness and greyscale.

## Structure

| Was | Is |
| --- | --- |
| Six alternating feature rows, ~4,000px | One ARIA tab set over one visual |
| Marquee of payment-rail names | "Waves never holds, moves, or touches your money" |
| Four-up stats bar (`0 ms`, `1 tap`) | Gone — the real numbers live in the sentences they prove |
| "180+ currencies" as a claim | A searchable, region-filtered list of every active ISO 4217 code |
| Four generic privacy bullets | What Waves can't see / does hold / doesn't protect against |
| "Early access", no number | "Not priced yet", said plainly, plus an honest tier table |
| Boxed FAQ cards | Hairline rows, oversized muted questions, sticky heading |
| 9,900px | 8,600px |

The settle-up claim, which the copy has always made and never shown, is now drawn: seven debts tangled between five people, beside the same group cleared by two payments.

## Not India-first

The hero card, every drawn visual and every worked example are **localised per language** rather than translated from one currency: euros with a fixed pound rate in English, rupees in Tamil and Hindi, dirhams in Arabic — with the arithmetic consistent in each (the split visual's four shares add to its total in all four). `priceCurrency` follows the locale. Free is the word "free", not an amount in somebody else's money. Payment rails are examples in no privileged order, drawn from four continents.

## Accessibility

Audited against **WCAG 2.2 AA** with every contrast figure computed over the real composited backdrop, then verified with **axe-core across four locales at two widths**.

- The muted ink ramp now clears 4.5:1 on the page ground *and* on chip fills — the old ramp bottomed out at 1.46:1.
- The marquee's unstoppable infinite animation (2.2.2) is gone with the marquee.
- The mobile menu is a real `<dialog>`, so the focus trap, Escape and background inertness come from the browser rather than from our own listeners.
- The language popup is a disclosure, not a fake `role="menu"`, and `lang` sits on the endonym alone — a screen reader no longer reads "Tamil" in a Tamil voice.
- The skip link targets a focusable `<main>`; scrollable regions are keyboard-reachable and named.
- Drawn visuals are single `role="img"` nodes with translated labels instead of a hundred loose fragments read aloud between the real copy.
- `forced-colors` mode gets explicit borders; `body { overflow-x: hidden }` is gone, so the next reflow bug is visible instead of silently eating Tamil.
- An unmatched URL renders our own 404 inside the locale layout instead of Next's unstyled English fallback — and does it **without a dynamic API**, so every page stays prerendered.

**One axe violation remains, on every page, by design:** the giant watermark wordmark in the footer. WCAG 1.4.3 exempts text that is part of a brand name, which is exactly what it is; at 3:1 it would stop being a watermark. It is `aria-hidden` and carries no information.

## Being read by machines

Every page has a Markdown twin at `/{locale}/index.md`, generated from the same dictionary the page renders from so it cannot drift. It is advertised three ways: a `<link rel="alternate" type="text/markdown">`, an HTTP `Link:` header, and `Accept: text/markdown` negotiation decided on q-values only — never on the user agent, which would be cloaking. `/llms.txt` indexes it.

`middleware.ts` became `proxy.ts` per Next 16's deprecation, and now sets `Vary: Accept-Language` on the locale redirect and filters `q=0` refusals out of language negotiation.

## Deliberately not done

- **No invented proof.** No ratings, user counts, testimonials, press logos or `aggregateRating` — none of it is real yet. The slots exist; `NEXT_PUBLIC_IOS_URL` / `NEXT_PUBLIC_ANDROID_URL` turn the store links on when the listings do.
- **No App Store / Play badge artwork.** Both vendors require their own asset unaltered; a hand-drawn lookalike breaches their guidelines.
- **No comparison table against Splitwise.** It would need claims about a competitor that somebody has to verify.

## Gates

`typecheck`, `lint` and `build` clean. All 16 pages prerendered.

## Summary by CodeRabbit

- **New Features**
  - Added light, dark, and system theme selection.
  - Added searchable currency browsing across 180+ currencies.
  - Added localized Markdown pages and an `llms.txt` resource for machine-readable content.
  - Added localized 404 pages and app-store link support.
  - Added accessible feature tabs, improved navigation, skip links, and keyboard interactions.

- **Improvements**
  - Refreshed the site with a light-first ledger design, new typography, illustrations, pricing comparisons, and updated content.
  - Improved localized metadata, sharing previews, currency information, and language-aware page handling.


dmadan86:
Went through all seven CodeRabbit findings. Five were fixed in e9799ef; here are the two that were not, and why.

| Finding | Verdict | Reason |
| --- | --- | --- |
| `en.json:322-323` — worked settlement is impossible (€84 groceries, €62 electricity, €27 taxi, "You pay Tomás €5.50") | **Already fixed** | That copy was in 60549f7, the commit the review ran against. The branch was force-pushed since; on the current head the example is €60.00 groceries (You paid), €42.00 electricity (Tomás), €18.00 internet (Mei), all split four ways — so You are +30, Tomás +12, Mei −12, Nadia −30, and the two payments shown (Nadia pays you 30.00, Mei pays Tomás 12.00) settle it exactly. The Tamil, Hindi and Arabic examples were re-checked the same way and all three balance. |
| `markdown.ts:60` — add a localised "every active ISO 4217 code" claim | **Fixed differently** | The line is now `t.currencies.count` with the derived count, and the ISO 4217 claim is not re-stated: `t.currencies.subtitle`, emitted two lines above it, already makes that claim in all four languages. Inventing a second translated sentence for a claim already on the page would have been duplication. The other two strings on that finding (settlement rail label, footer contact label) are now dictionary values. |

Also removed `website/a+b` and `website/owed'` — two empty files a mis-quoted shell command had committed. They are what the `typecheck + lint` check was failing on, alongside the two checked-in `.woff2` faces, which `scripts/check-filenames.sh` does not yet recognise.

PR #724, opened 2026-09-08
The receipt stays on screen after it uploads, and has something to show before it arrives

dmadan86:
The report: *"it loads and gets uploaded, and there is a blank screen, there's nothing. It goes back to nothing, and then after some time, the image comes back, maybe downloading again."* Two separate problems, both fixed here.

## 1. The blank is a real bug

`runFlush` deleted the queue entry and its bytes the instant the attach RPC returned (`apps/mobile/src/lib/receiptQueue.ts:547` before this change). The attachment row that replaces it does not exist on this device until a sync pull brings it down — a network round trip later — so for that whole window the gallery had **neither** a pending entry nor a row, and the strip went empty. The row then arrived under a new key and `useResolvedUrls` started from `undefined`, so the picture reappeared as if it had been fetched. The de-dupe in `ExpenseReceipts` written for exactly this handover could never fire, because the thing it was handing over *from* was deleted first.

- A successful upload now marks the entry `sentAt` and **keeps** it, still drawing the same picture. The gallery drops it when the row it became lands beside it (`dropSettledReceipts`); an hour-long TTL sweeps the case where the pull never comes.
- The settled entry resolves to the **view cache under the key the real row resolves to**, so the swap from entry to row is the same file — nothing reloads, nothing crossfades.
- Cached/local images now resolve **during render** rather than from an effect, so an image already on disk never shows a spinner for a frame first.
- Parked captures sort **with** the attachment rows (newest first) instead of after them, so a new tile no longer jumps from the end of the strip to the front when the sync catches up.
- Removing a settled capture goes through the real removal RPC — it is on the server, and dropping the local entry alone would have orphaned it.

## 2. A low-quality placeholder, on both screens

A ~32px JPEG of the receipt now travels inside the row as a `data:` URI (under a kilobyte — cheaper than the request it covers for), drawn under the real image and cross-faded out when it decodes. Both the edit screen and the view-expense screen share `ExpenseReceipts`, so both get it, on the thumbnail strip and in the full-screen viewer.

Built on `expo-image`'s own `placeholder` + `transition` and `expo-image-manipulator`, both already dependencies — **no new native module**, so this still ships over OTA once the migration is deployed. The placeholder is the same `Image`'s placeholder, not a second image, so a screen reader is never told there are two pictures. Tiles are fixed-size and the viewer is full-screen, so nothing reflows.

No new user-visible strings, so no i18n changes.

## Deploy-gated

`20260908160000_receipt_preview_placeholder` adds `expense_attachments.preview`, a 4 KB check constraint, a `waves_clean_image_preview` validator, and recreates `waves_attach_expense_attachment` / `waves_replace_expense_attachment_image` with a defaulted `p_preview`. Dropped-and-recreated rather than overloaded: a second signature differing only by a trailing default makes every existing four-argument call ambiguous through PostgREST. Old clients omit the argument and store no preview.

`pnpm db:migrate` is the whole deploy — the `sync` edge function selects `expense_attachments` as `*`, so the new column reaches devices with no edge change, and neither `api-client` nor `apps/web` carries attachments. Until the migration lands, previews are simply never stored and every screen behaves as it does today; the blank-flash fix is client-side and independent of it.

## Verified

- `packages/db`: full suite, 766 tests, against a database rebuilt from every migration in order (including this one, and the anon-surface guard that counts reachable definer functions). New `test/receipt-preview.test.ts` covers the kept/dropped/oversized/`javascript:` cases, replace-overwrites-preview, and that a four-argument client still attaches.
- `apps/mobile`: 977 tests, including six new ones on the handover — the entry survives its own success, resolves to the cache file, is never re-sent, is let go when the row lands, is reaped when it never does, and keeps its local bytes when the cache write fails.
- `packages/core`: 922 tests. `tsc --noEmit` clean on mobile and core; `expo lint` clean; `prettier --check` clean.

**Not verified without a device:** the actual look of the crossfade and of a 32px placeholder upscaled to a 96pt tile and to full screen, the real timing of the pending→settled→row handover on a phone, and that `expo-image-manipulator` produces a preview under 4 KB from a real camera photo on both platforms (it falls back to no preview if not).

## Summary by CodeRabbit

* **New Features**
  * Added blurred image previews that appear while receipt images load.
  * Receipt uploads now remain visible locally until synchronization completes.
  * Pending receipts are prioritized and sorted newest first.
  * Users can remove receipts that have already synced with the server.

* **Bug Fixes**
  * Improved receipt loading by using local images and cached previews when available.
  * Prevented duplicate or stale queued receipts from remaining after synchronization.
  * Preserved local receipt images even if caching fails.

### DIFF
--- a/apps/mobile/src/components/ExpenseReceipts.tsx
+++ b/apps/mobile/src/components/ExpenseReceipts.tsx
@@ -58,6 +58,7 @@ import { imageUrl, restrictedImageUrl } from '@/lib/storage';
 import { cacheImage, cachedImageUri, evictImage } from '@/lib/storage/imageCache';
 import {
   discardPendingReceipt,
+  dropSettledReceipts,
   enqueueReceipt,
   flushReceiptQueue,
   pendingReceiptUri,
@@ -80,6 +81,31 @@ type GalleryItem =
   | { kind: 'pending'; key: string; entry: PendingReceiptView };
 
 /**
+ * The file an item can be drawn from *right now*, with nothing awaited: a
+ * capture the queue is still carrying is its own local file, and a bill seen
+ * before is on disk under its stable path (ADR-005). Null when only a
+ * short-lived signed URL will do.
+ *
+ * Cheap enough to ask on every render — it is a stat — which is the point:
+ * answering during render rather than from an effect is what stops an image the
+ * device already holds from showing a spinner for a frame first.
+ */
+function localUri(it: GalleryItem): string | null {
+  if (it.kind === 'pending') return pendingReceiptUri(it.entry);
+  const bucket = it.kind === 'legacy' ? 'receipts' : 'expense-attachments';
+  const path = it.kind === 'legacy' ? it.path : it.row.storagePath;
+  return cachedImageUri(bucket, path);
+}
+
+/** The tiny stored stand-in for an item's image, when it has one. */
+function previewOf(it: GalleryItem): string | null {
+  if (it.kind === 'attachment') return it.row.preview;
+  if (it.kind === 'pending') return it.entry.preview ?? null;
+  // The legacy bill predates the column; there is nothing to have kept.
+  return null;
+}
+
+/**
  * Resolve every item to a displayable URL, each through its own backend (the
  * legacy bill is group-readable in the `receipts` bucket; an attachment is
  * signed by subject in the restricted bucket). Keyed by item so a resolved URL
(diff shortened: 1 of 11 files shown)


dmadan86:
CodeRabbit's four findings, verified one at a time — two taken, one taken in a narrower form, one rejected on evidence.

**1. Serialize the `QUEUE_KEY` read-modify-writes — taken.** Real and reachable. Each mutation reads the index and writes a derived one, both asynchronously; two that overlap both read the same snapshot and the later write discards the earlier change. Most pairs are cosmetic, but one loses a receipt: an add whose resize takes a second, landing while a flush or the gallery's handover writes back a queue that never saw it — the bytes then sit under the pending dir with nothing pointing at them, for the next orphan sweep to delete. The hazard pre-dates this branch; the branch added two more writers firing from a screen effect, which makes it materially easier to hit.

Every mutation now takes a turn on a module-level chain. Uploads and the flush's network work stay outside it, so a slow send never blocks an add. Two things moved *inside* a turn that were not atomic before, and both matter: an enqueue's bytes and the entry naming them (split across turns, the orphan sweep is entitled to delete a photograph mid-add), and the sweep itself. Two new tests cover it, and I confirmed both fail with the chain stubbed out to a pass-through.

**2. The TTL sweep never runs on its own — taken, narrowed.** The call graph holds: `flushReceiptQueue` runs from the sync provider on launch, foreground and network-connect, and from `ExpenseReceipts` on mount and on connection change — but the periodic sync timer calls `syncEngine.flush`, not this. So a gallery left open for over an hour triggers nothing. The consequence is mild (the tile shows a receipt that really is on the server), so rather than add a timer for it, the rule now lives in one exported predicate that the gallery applies to what it *draws* and to what its handover effect lets go of. That closes every case except a screen that does not re-render at all for an hour, which is stated in the code rather than papered over.

**3. No preview when source dimensions are unavailable — taken.** Worth distinguishing two causes. The scanner path reaches zero dimensions precisely *because* `imageSize` already failed on that URI, so asking again returns the same nothing — the suggestion is a no-op there. But the picker can also simply omit dimensions for a file `Image.getSize` reads perfectly well, and that case was giving up for no reason. `previewDataUri` is fully guarded (it returns null with no file access when the manipulator is missing, and catches everything else), so asking costs nothing on the failing path and gains a placeholder on the other.

**4. Add the CHECK as `NOT VALID` — rejected.** The premise about the scan is right — Postgres does run a validating scan for `ADD CONSTRAINT ... CHECK`, and it does not skip it just because the column was added in the same migration and is all-NULL. But the conclusion does not follow here: `expense_attachments` on prod has **1 row**, 64 kB total (this project's database was rebuilt on 2026-09-04). An ACCESS EXCLUSIVE lock held for a one-row seq scan is not worth buying off with a constraint left permanently `convalidated = false` that someone has to remember to validate in a later migration. If this table grows, a future constraint on it should absolutely be `NOT VALID`; this one should not.

Gates after the changes: mobile 981 tests, core 922, `tsc --noEmit` clean, `expo lint` clean (3 pre-existing warnings in an untouched file), `prettier --check` clean. `packages/db` is unchanged this round; its 766 tests passed earlier against a database rebuilt from every migration in order.

Still needs a device: unchanged from the original description, plus the turn-taking chain's behaviour under a real slow file copy — the tests drive it through a fake filesystem, so they prove the ordering, not the timing.

PR #749, opened 2026-09-09
Answer questions in the app's own voice, not the phone's

dmadan86:
Every confirmation in the mobile app was Android's stock `Alert.alert`: a grey slab with square corners, a left-aligned white title, a wall of body text, and two teal ALL-CAPS text buttons — `CANCEL` / `DELETE ANYWAY` — with the destructive one on the right in the same colour and weight as the safe one. It sat over a screen of rounded cards, brand purple and tinted danger rows, and it belonged to a different application.

It could also be told nothing but strings, which is how the delete-group warning came to render four real debts — Lokesh owes Madan ₹15,171.66, Renny ₹4,996.79, Gayathri ₹3,602.49, Hethu ₹411.48 — as one undifferentiated paragraph. Money in a list should look like money in a list.

All **49 call sites across 24 files** are gone. Nothing in `apps/mobile` imports `Alert` any more, and an eslint rule stops the fiftieth.

## What Mobbin decided

**Weight follows consequence.** The irreversible door is a filled red button and it comes first; the way out is the soft full-width one under the thumb. That is [Nextdoor "Delete post?"](https://mobbin.com/screens/59bce3a8-29e0-4af4-ba3a-d3f5555a1b94), [Peerspace "Delete this board"](https://mobbin.com/screens/2b2bdae0-6d45-44ab-8bdf-bcb5516480cc), [Instacart "Remove a member"](https://mobbin.com/screens/6be4d6a0-d234-4356-9a67-0a2571996260) and [Lyft "delete this family"](https://mobbin.com/screens/eef1ecdd-472e-4683-bcdc-1f96193ba123) — and the hierarchy #747 already set for sign-out. The native alert's two same-coloured capitals are the thing this replaces.

**The door says what it does.** "Delete family" (Lyft), "Remove goal" ([Cash App](https://mobbin.com/screens/77421f0e-576e-43e6-9ef1-7e9a3c87f28b)), "Yes, delete this board" (Peerspace) — never "OK". A screen reader hears the verb and its object, which is the whole warning for somebody who cannot see the title.

**A destructive dialog gets a mark; an ordinary one does not.** [Notion "Delete transcript"](https://mobbin.com/screens/1dc2682d-9a06-41b9-86f9-28e2e5155ab8) and [Fabric "Delete action"](https://mobbin.com/screens/fc3a4134-df6e-487c-96c3-ee14b8352eff) draw theirs bare; Lyft, Cash App and [Too Good To Go](https://mobbin.com/screens/470f64da-21f4-4f5f-9fa9-40c35beb0255) put an alert glyph on the ones that end something. A warning icon on every dialog is a warning icon nobody reads.

**Centred card vs bottom sheet.** Notion and Fabric are two-button centred cards. Everything carrying a list, a paragraph of consequence, or a third option comes up from the bottom edge: Nextdoor, Instacart, Peerspace, Cash App, [Origin "Delete holding"](https://mobbin.com/screens/2a8bdb8a-1b26-48c4-affd-e73fba533d14). So the app picks by shape, not by caller — see the API below.

**Consequences are shown, not summarised.** Cash App's "Remove goal" lists what happens as rows. [Acorns "Close your Invest account"](https://mobbin.com/screens/b3fa6fd2-dc44-4887-bc2e-4c875afe2550) lists the balances with amounts aligned to the right of each label; [Stake](https://mobbin.com/screens/c1330f0a-fd78-49c5-97f8-2e358cfe2e41) and [PayPal "Review and convert"](https://mobbin.com/screens/e2fb9e2e-fdb9-45ea-bd67-bcd00f6f58fb) do the same for a transfer. That is exactly the delete-group case, and it is why the API takes rows.

**The last word gets a panel, not a fourth paragraph.** [Wise "Closing your balance"](https://mobbin.com/screens/c7da1850-c80b-4096-837f-5d96daf315fa) puts the consequence in a tinted callout rather than a run of grey sentences. Our `note` field does that with the existing `Callout`.

## The API, and why

A provider at the root plus a hook returning a promise:

```ts
if (await confirm({ title, body, confirmLabel, tone: 'danger' })) remove();
const picked = await choose({ title, options });   // id, or null
await notify({ title, body });                     // one door
```

**Why not per-site state.** `Alert.alert` is imperative; `Popup` is declarative. Bridging at the call site means every screen that asks a question grows a `useState`, a second piece of state for whatever the answer applies to, and a block of JSX at the bottom of its tree. Across 49 sites that is not a refactor, it is a rewrite of 24 screens into overlay hosts, and the noise lands in the screens that are already the busiest. Mounted once at the root it is *shorter* than the alert it replaced, works from a `catch` block, and survives the asking screen being popped — the same reasoning that already puts `ToastProvider` at the root.

**Surface is chosen by shape, not by caller** (`presentationFor`): rows present, or more than two actions, or a body over 220 characters → `Sheet`; otherwise `Popup`. So the same kind of question never turns up centred on one screen and bottom-anchored on the next. It matters mechanically too — a sheet is anchored to an edge and can be given a definite height to scroll inside; a centred card that outgrows the window has nowhere to put the overflow.

**Three call shapes, one surface, three button grammars.** `confirm` → filled (red or brand) + soft full-width. `notify` → one soft button; a lone filled primary on a notice reads as a decision being asked for. `choose` → soft peer rows, the destructive one red-on-soft, the way out a ghost — the action-sheet grammar, because a filled brand button on the top row turns a menu into a recommendation nobody asked it to make.

**Where it lives.** `AppDialog` is in `apps/mobile/src/components`, not `@waves/ui`: it needs the app's words (the default cancel label, the scrim's spoken close label, `t.common.ok`), and a design-system component taking eleven strings as props would push that decision back out to the 49 call sites this exists to shorten. Everything it is *made of* — `Sheet`, `Popup`, `Button`, `Callout`, `MoneyText`, `Row` — stays in the design system. No raw `Modal` anywhere.

## The dismiss and race contract

In `apps/mobile/src/lib/dialogQueue.ts`, pure, tested in `apps/mobile/test/dialogQueue.test.ts` (13 cases):

- **Every request settles, exactly once.** A door, a scrim tap, Android's back gesture (through `Modal`'s `onRequestClose`, which both `Sheet` and `Popup` already wire), or a refusal from a full queue. A dismissal resolves `null`; `confirm` reads that as no. Nothing can leave a caller `await`ing forever.
- **One at a time, the first one wins.** A second `confirm` while one is open **waits behind it** rather than replacing it — a question that swapped itself out mid-read is how somebody taps "Delete" meaning "Cancel". Past `MAX_WAITING_DIALOGS` (3) waiting, the **newest** is refused and resolved as a dismissal, so a failing loop cannot build a wall of dialogs; the oldest is kept because it is the one somebody actually asked for.
- **A stale answer is ignored.** `closeDialog` refuses any id that is not the one on screen, so a double tap or a late close is silent rather than resolving a promise twice or skipping a question nobody answered.
- **The asking screen may unmount.** The surface lives above the navigation stack, so the dialog is unaffected; the caller's continuation runs on a possibly-unmounted component, which is the same hazard `Alert.alert`'s callbacks had.

Each surface remembers the last request *it* drew, separately. Choosing "Delete" from an action sheet immediately asks for a confirmation, so a sheet is sliding down while a card fades in — with one shared memory the departing sheet would spend its last frames wearing the incoming dialog's words.

## The 49 sites

| Kind | Count | What they are |
|---|---|---|
| **Confirmations** (`confirm`) | 28 | delete group / expense / capture / budget / loan / recurring / tag / comment / receipt / settlement proof, leave group, archive group, block a person, merge people, uninstall a pack, cancel or reject a settlement, confirm all settlements, collapse several payers, record a payment, receipt-cap upsell, restart for RTL |
| **Toasts** (`useToast`, negative) | 13 | one-line failures with nothing to decide |
| **Action lists** (`choose`) | 4 | profile photo menu, comment ··· menu, add-receipt (scan/library), unsent-receipt (retry/discard) |
| **Notices** (`notify`) | 4 | settle-up first before leaving, no payment details for this rail, batch-assign partial result, restart required (build cannot self-restart) |

### Behaviour that genuinely changed

**13 blocking alerts became non-blocking toasts.** These were all `Alert.alert(oneSentence)` with a single implicit OK. Taking the screen to say "could not save" and then demanding a tap to give it back is a tax on a failure that was not the person's fault — and in every one of these the screen itself is already the message (the switch flipped back, the tile is still there, the receipt is still open).

- `receipt/[id].tsx` — could not save image
- `settings/privacy.tsx` — could not save consent (the toggle has already reverted)
- `ExpenseComments.tsx` ×3 — could not post (add), could not post (edit), could not delete
- `ExpenseReceipts.tsx` ×5 — could not keep bytes, could not remove, could not save image, could not save annotation, could not save adjustment
- `captures.tsx` ×2 — batch-assign threw, new-people-group threw
- `group/[id]/settings.tsx` — delete refused (`NOT_ADMIN` or unknown). **Worth a look in review:** this one fires straight after the person dismissed the delete confirmation, and stacking a second dialog on the one just dismissed is how somebody taps through a sentence without reading it. It is a red toast instead.

`ToastProvider.show` gained an optional `tone` for this (positive still the default).

**Two dialogs became quieter on purpose.** *Archive group* dropped its danger mark and red door: archiving puts a group away, it does not throw anything out, and it can be undone from the archive — a distinction the native alert could not draw. *Confirm all settlements* likewise: accepting money owed to you is not destructive.

**`LanguageProvider` no longer owns UI.** It is the outermost provider in the tree, above the theme and above `DialogProvider`, so it could not use any of this — it used to reach past all of it for `Alert.alert`. It now raises a `restartPrompt` value and a new `LanguageRestartPrompt` component (mounted inside `DialogProvider`) draws it. The words still come from the language *just chosen*, which is the one the person has said they read.

## The delete-group dialog

The one this whole change was for. `groupDeleteBody` (which joined everything into one string, because the native alert accepted nothing else) is now `groupDeleteWarning`, returning `{ body, rows, moreRows, note }`. The screen hands over `{ label: "Asha owes Ravi", amount: { minor, currency } }` per debt and `MoneyText` draws the amount — money colours, faded minor units, the spoken label money carries everywhere else. Still the biggest four, still counted after that, still every currency, still plain member names rather than "You". Because it carries rows it is a bottom sheet, and the scroll region is a direct child with `flexGrow: 0` and a **definite `maxHeight` in points** (the #747 gotcha).

The other 48 were checked for the same shape; none of the rest carries per-row money.

## i18n

One key changed: `group.deleteOwesLine` (`{from} owes {to} {amount}`) → `group.deleteOwesWho` (`{from} owes {to}`), since the amount is now drawn beside the label instead of formatted into the sentence. Translated by hand in all four tables — en/ta/hi/ar — and the doc comment above the interface updated.

Everything else reuses the strings the call sites already had; no site lost its translation and no new English was written. RTL throughout: `Row`, no physical margins, no hardcoded left/right.

## Accessibility

- The content is wrapped in `accessibilityRole="alert"` named by the title; `accessibilityViewIsModal` comes from the surface, and a native `Modal` gives Android its own window.
- The scrim's spoken label is the dialog's *own* way out ("Cancel", "Not now", "Keep it"), not a generic "Close".
- Each consequence row is `accessible` as one item — the name and the amount are one fact, and hearing them as two lines loses which belongs to which when there are four.
- Destructive labels name what they delete; there is no "OK" on a destructive door anywhere in the change.
- Android hardware back dismisses and resolves as cancel, on both surfaces.

## The guard

`apps/mobile/eslint.config.js` gains a `no-restricted-imports` entry banning `Alert` from `react-native` — the same mechanism already guarding the Supabase boundary and the raw expo-router. Verified it fires.

```
$ grep -rn "Alert.alert" apps/mobile/src
(no matches)
```

## Gates

Windows node, from the worktree.

| | |
|---|---|
| `pnpm format` → `pnpm format:check` | pass — "All matched files use Prettier code style!" |
| `pnpm lint` (all projects, `--max-warnings 0`) | pass, 0 problems |
| `pnpm typecheck` (mobile) | pass |
| `pnpm test` — mobile | **90 files, 1206 tests** pass (13 new in `dialogQueue.test.ts`, `groupDeleteWarning.test.ts` rewritten for the structured return) |
| core / web / api / admin / api-client / agent-mcp | pass |
| `packages/db` | fails on connection setup — needs a live Postgres on `localhost:54330` and a `.env` this worktree does not have. Unrelated. |

## Deliberately left alone

- `apps/web` and `apps/admin` — no `Alert` there to begin with.
- The four `notify` sites stayed dialogs rather than becoming toasts: each explains why an action was *refused*, and the next thing to do is act on that explanation.
- `captures.tsx`'s batch-assign summary stayed a dialog because it reports a partial run ("2 saved, 1 could not be") — a fading line is the wrong shape for a result somebody has to reconcile.
- Nothing device-tested. Every gate is JS-only; the motion, the sheet heights and the screen-reader order want a real phone.

## Summary by CodeRabbit

* **New Features**
  * Added a unified in-app dialog system for confirmations, notifications, and selection menus.
  * Replaced native alerts across profile, groups, expenses, comments, receipts, settings, and personal finance screens.
  * Added structured group-deletion warnings with itemized debts and clearer details.
  * Added language-change prompts with restart guidance.
  * Added positive and negative toast notification styles.

* **Bug Fixes**
  * Confirmation actions now proceed only after explicit user approval.
  * Dismissed dialogs consistently cancel pending actions.

* **Tests**
  * Added coverage for dialog queuing, presentation, dismissal, and group-deletion warnings.


dmadan86:
## Review fixes

All 11 confirmed defects fixed, plus 5 of the 8 plausible. Gates green: `format` → `format:check` → `lint` (all projects, 0 problems) → `typecheck` → `test` (mobile **1217** tests, was 1206; core/web/api/admin also pass; `packages/db` still needs a live Postgres).

### Confirmed

**1 — five invisible Toasts.** Confirmed exactly as described. Fixed by saying each message inside the modal that raised it, rather than moving the toast host (there is nowhere above a native window for an in-tree view to go):

- `ExpenseComments` composer → an inline `Callout` above the toolbar. Cleared when the composer opens and on each fresh send, so a retry does not read as the previous failure.
- The receipt viewer, the annotator and the cropper → a new `components/ModalNotice.tsx`: a floating, **non-fading** panel at the foot of a dark full-bleed modal, dismissed by hand. The annotator and cropper take `error` / `onDismissError` props, since the mutations that fail live in the parent. The viewer's notice takes an `offset` so it clears the page-counter pill.

**2 — `couldNotKeep`.** Agreed, and the file said so itself. Back to a dialog (`notify`), with the reasoning written next to the comment that was already there. `settings.tsx`'s `NOT_ADMIN` took your suggestion: a persistent `Callout` under the delete row, which satisfies both the don't-stack-on-a-just-dismissed-dialog point and the don't-fade-over-an-unchanged-screen point.

**3 — scrim bound to `current`.** Confirmed. `AppDialog` now takes a `DialogSlot` per surface and `onChoose(id, choice)`; every route out — doors and scrim — carries the id of the dialog **it was drawn for**, so the second tap arrives stale and the hub drops it. Covered by a test.

**4 — swap in place.** Confirmed. A `HANDOFF_MS` (190 ms, past `CLOSE_MS`) gap between one dialog closing and the next opening, so the first plays its exit, the second plays its entrance and announces itself. This also closes **P2**: the cross-surface exit race cannot happen when the surfaces never overlap.

**5 — the restart prompt asks twice.** Confirmed; the comment described behaviour the code did not have. It now suppresses a second prompt while one is open — the second is word for word the first, so there is nothing it could add — and the comment says what actually happens.

**6 — never announced.** Confirmed: `accessibilityRole="alert"` on a View with no `accessible` is inert, and adding `accessible` would have swallowed the doors on iOS. It now speaks via `AccessibilityInfo.announceForAccessibility` (title, then body) keyed on the request, plus `accessibilityLiveRegion="assertive"` for Android. Keyed on the request object, so a promoted dialog announces rather than silently replacing the words.

**7 — silent outside the provider.** `__DEV__` warn added, naming what will happen.

**8 — unmount strands promises.** Confirmed. `dispose()` settles everything still waiting as `DIALOG_UNASKED`, called from a `useEffect` cleanup. Two tests.

**9 — refusal ≡ cancel, and a rationed `notify`.** Both fixed:

- `DialogRequest.mustBeSeen` exempts a request from the depth cap; `notify` sets it. A report is not a question.
- Refusal (and disposal) now resolve `DIALOG_UNASKED`, not `null`. `confirm` still reads it as false — documented — because a question nobody saw is certainly not a yes. `friends/merge` is the one site whose *no* moves somebody, so it asks through `ask` and checks `isUnasked`; the comment says why it is the exception.

**10 — notify's scrim said "Close".** Fixed at the source: `notify`'s single door now carries `DIALOG_CANCEL`, because it *is* the way out, so `labelFor` finds it.

**11 — TDR drift.** `waves-tdr.md` A64 updated: `deleteOwesWho` rows with the amount drawn beside them as money.

### Plausible

- **P1 — accessible scrim hides the card.** Agreed and fixed in `@waves/ui`. Both `Sheet` and `Popup` now put the scrim `Pressable` as an `absoluteFill` **sibling** under a `box-none` frame holding the card, so a screen reader finds the title, rows and doors instead of one button called "Close". Documented at the top of `Overlay.tsx` as a rule for both surfaces.
- **P4 — heights.** Agreed on both halves. `Popup` now gets a `maxHeight` (80% of the window; the sheet's 88% is now named alongside it), and `DialogContent`'s root carries `flexShrink: 1` so the surface's `maxHeight` can actually squeeze the scroll region instead of the doors being pushed past the card.
- **P5 — `require` form.** Agreed. Two `no-restricted-syntax` selectors now catch `Alert.alert(…)` and `X.Alert.alert(…)` however `Alert` was obtained. Verified against both `const { Alert } = require('react-native')` and `require('react-native').Alert.alert`.
- **P6 — `id: 'cancel'` collision.** Agreed. `choose` namespaces caller ids with an `option:` prefix internally and strips it on return, so a caller may name an option anything.
- **P7 — re-render on every flip.** Agreed. `restartPrompt` moved off `LanguageValue` onto its own context consumed only by `LanguageRestartPrompt`.

### Dismissed

- **P8 — `AppDialog` never clears `sheetSlot` / `popupSlot`.** Not fixing. It retains at most two request objects for the life of the process, and `Sheet`/`Popup` already render nothing once their mount latch drops — there is no leak and nothing drawn. Clearing them would need a "finished closing" callback `Overlay` does not have, and adding one to the design system to free two objects is not a trade worth making.
- **P3 — a root Modal raised from inside a presented Modal.** Not guessing; flagged for device verification, as you suggested. Fix 1 removed the five toast-under-modal cases, but six `confirm`/`choose` sites still raise the root dialog from inside a presented modal (the receipt viewer's remove, the settlement-proof viewer, the tag editor inside a Sheet, the comment action sheet). Each was `Alert.alert` before, so each is a real change in stacking behaviour. **This is the one thing in the PR that a reviewer should check on a device before merge.**

### Also noted

You are right that `ExpenseComments.openActions` can reach four buttons where Android's `Alert.alert` silently truncated to three — that is now a bottom sheet with four full-width rows and no truncation, which is a fix rather than a regression, but it is a visible behaviour change and worth a look on a device.

### Test gap

Closed. The queue, the promises and the settling rules are now a `createDialogHub` in `lib/dialogQueue.ts` — the provider is a thin React wrapper over it. Eleven new tests cover exactly what was uncovered: resolving with the door taken, dismissal as a no, two questions each answered with their own door, a stale second answer settling nothing, refusal as `DIALOG_UNASKED` with nothing displaced, `mustBeSeen` never refused, and `dispose` settling everything open (plus asking after disposal).

PR #771, opened 2026-09-12
Follow-up: serialize phone OTP relay exchanges

dmadan86:
Follow-up to merged PR #769.\n\n- Refuses a second live Firebase→GoTrue phone relay exchange for the same number instead of replacing the first row.\n- Keeps stale relay replacement so abandoned exchanges do not pin a number forever.\n- Makes phone-verify's gate refusal copy neutral for cap changes and blocked-number cases.\n- Strengthens phone-verify tests to assert relay-open precedes the GoTrue OTP request on one shared timeline.\n- Adds static SQL coverage for relay ownership/concurrency because local Postgres is not always available in audit environments.\n\nValidation:\n- pnpm test:edge -- phone-verify otp-send\n- pnpm --dir packages/db exec vitest run test/otpRelaySql.test.ts --pool=forks\n- pnpm --filter @waves/db typecheck\n- pnpm typecheck\n- pnpm lint\n- bash scripts/check-filenames.sh\n- git diff --check\n\nNote: running the package-level DB Vitest command pulled in integration tests and failed because local Postgres on localhost:54330 is unavailable; the targeted static SQL test passed directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved phone verification handling to prevent multiple simultaneous OTP exchanges for the same number.
  * Updated rate-limit messaging to clearly state that too many attempts were made for the number and to try again later.
  * Preserved the existing rate-limit response behavior, including the 429 status and error code.

### DIFF
--- a/packages/db/prisma/migrations/20260912150000_refuse_concurrent_otp_relay/migration.sql
+++ b/packages/db/prisma/migrations/20260912150000_refuse_concurrent_otp_relay/migration.sql
@@ -1,0 +1,32 @@
+-- A second live Firebase→GoTrue phone exchange for the same number must not
+-- replace the first one. Replacement lets one device claim or close another
+-- device's code, and can make the newer request fall through to ordinary SMS
+-- delivery. Stale rows are still replaced so an abandoned exchange does not pin
+-- the number forever.
+
+CREATE OR REPLACE FUNCTION public.waves_otp_relay_open(p_phone text)
+RETURNS uuid
+LANGUAGE plpgsql SECURITY DEFINER
+SET search_path TO 'public', 'pg_temp'
+AS $$
+DECLARE
+  v_exchange uuid := gen_random_uuid();
+BEGIN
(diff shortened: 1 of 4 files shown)


dmadan86:
Superseded by #770, which is now on main.

Every change proposed here already landed: the relay busy guard (`20260912160000_refuse_concurrent_phone_relay`, with stale-row replacement kept so an abandoned exchange cannot pin a number), the neutral gate refusal copy, the shared-timeline assertion that relay-open precedes the GoTrue request, and `packages/db/test/phoneRelaySql.test.ts`.

The same change was also pushed directly onto `feat/phone-sign-in-firebase` as `c764c04e`, so it reached main by that route. Two tests written against the previous displacement behaviour were reconciled with it there, and the migration is applied to the live project — `waves_otp_relay_open` answers a second caller `OTP_RELAY_BUSY`, verified against it.

Closing rather than merging: the branch conflicts with main and resolving it would produce no change.

PR #779, opened 2026-09-12
Research: what Waves can notify and mail, and where each one should run

dmadan86:
`docs/notifications-and-email-strategy.md` — one document, no code. It reads the
whole path (`waves_notify`, both claim functions, `notify-fanout`, the Resend
half, the device side, the settings screens) and writes down what is true today
before proposing anything.

## The findings

**Most of the notification system is written and delivered to nobody.** Twenty-two
kinds have copy in four languages. Eleven have a producer in SQL. Of those eleven,
six can reach a person only by email and **five reach nobody at all** —
`settlement_confirmed`, `settlement_cancelled`, `settlement_disputed`,
`expense_disputed`/`_resolved`, the three ghost-claim kinds and both trip nudges
are written to `notifications`, closed out as terminally failed by the push claim
(no live token is a *decision*, not a retry), never mailed (not on the email
whitelist), and never seen — the in-app inbox went in #565.

**Push has never delivered a message on this project.** `push_tokens` has been
empty since the fanout was first scheduled; there is no `google-services.json` in
the checkout; the iOS APNs key is recorded as not done. Five things must all hold
for a push to land and at least two of them do not.

**The most obvious notification in a splitting app does not exist.**
`expense_added`, `expense_edited`, `expense_deleted`, `you_owe`,
`group_invite_accepted`, `ghost_claimed`, and — with *email templates already
written* — `settlement_initiated` and `digest_daily` have copy, a preference
switch that governs them, a renderer and tests, and **no producer anywhere**.

**Not one local notification is scheduled.** No `scheduleNotificationAsync` in the
repo; `expo-notifications` is used only for permission, token and tap routing.

**What actually reaches a person today** is five emails — new-device sign-in,
settle-up confirm request, added-to-a-group, a nudge, and the opt-in weekly digest
— and nothing else. None of them reach a phone-OTP account, because
`waves_email_for` requires a confirmed address.

## The split

The rule applied throughout: **anything derived from the private personal ledger
is scheduled on the device and never becomes a `notifications` row** —
`title`/`body` are plaintext columns, and a server job would have to read inside a
blob the server exists not to read. The cost (a reminder only lives on a phone
that has opened the app) is stated where it bites. Captures fall on the local side
too, for the same reason.

Ranked: settlement transitions (server, already built) · personal recurring due
(local) · trip-ended settle-up (server) · expense-added *involving me*, coalesced
(server) · invite accepted (server) · budget threshold (local) · trip starting
(server) · captures (local) · backup stale (local). Automatic dunning is
explicitly *not yet*.

## Email cadence

**Weekly by default, monthly offered, daily never.** Reasoning per type, not one
blanket answer: immediate for a settle-up confirm request and a security notice;
weekly/monthly for the balance summary; once-after-`end_date` for a trip recap;
the group digest folded into the summary rather than sent separately; and the
month-end personal summary refused outright as email because it is the private
ledger. Each proposal is checked against the two-file kind whitelist
(`TEMPLATE_FOR_KIND` + the `n.kind IN (…)` list in
`waves_claim_email_notifications`) and against the `factsOf` placeholder
whitelist, with the two that would need widening flagged.

Also covers: a settings screen that separates device-scheduled from server-sent;
quiet hours and the fact that `profiles` has no timezone column (only
`groups.time_zone`), with the rule that stops `waves_phone_gate`'s session-zone
bug recurring; existing and missing frequency caps; what push needs end to end;
and phasing marked JS vs deploy-gated.

## Open questions for you

1. Regular summary cadence — *Off / Weekly / Monthly*, weekly default, daily not
   offered at all. Accept?
2. `weeklyEmail` is a boolean today. Migrate it to a three-way (needs a
   read-both-shapes step or existing opt-ins are silently dropped), or keep the
   boolean and add a separate monthly toggle?
3. Local money reminders name the thing but **not the amount**, because a lock
   screen is the one place the private ledger is readable without the biometric
   gate. Opt-in "show amounts"?
4. Expense notifications: everyone in the group, or only people the expense
   involves? (Recommendation: involves-me.)
5. Coalescing window for expense activity — 15 minutes per recipient per group, or
   rely on involves-me alone?
6. Trip recap: push, email, or both?
7. Automatic dunning (`reminders.auto`, built and unfired) — wait for quiet hours,
   or ship at a fixed hour in the group's timezone?
8. Phone-only accounts get no email at all. Acceptable, or should phone signup ask
   for an address so the fallbacks work?
9. `digest_daily` and `settlement_initiated` are half-built — templates and
   whitelist entries, no producer. Delete or finish?
10. Per-group notification granularity — ADR-010 says considered and not built. A
    twelve-person trip is the case that wants it. Decide now or design the screen
    twice?

Two things to check on the live project that this document could not verify from
the repo: whether `cron.job` includes `waves-weekly-digest` (it self-schedules
from a migration guarded on `pg_cron` existing, and `pg_cron` is not installed by
default on a new project), and `select count(*) from push_tokens where revoked_at
is null`.

Not for merge until discussed.

### DIFF
--- a/docs/notifications-and-email-strategy.md
+++ b/docs/notifications-and-email-strategy.md
@@ -1,0 +1,793 @@
+# Notifications and email: what exists, what should, and where each one runs
+
+A research document, not a build. It answers four questions: what this app can
+produce today and which of it reaches a person; what belongs on the device; what
+belongs on the server; and what belongs in a mailbox, at what cadence.
+
+Nothing here contradicts `docs/plan-recurring-money-events.md` (merged) or
+ADR-010 / TDR §7. Where it goes further than either, it says so.
+
+---
+
+## 0. The three findings that matter
+
+1. **Most of the notification system is written and delivered to nobody.**
+   Twenty-two `kind`s have copy in four languages; eleven of them have a
+   producer in SQL; and of those eleven, **six can currently reach a person only
+   by email, and five reach nobody at all.** Push has never delivered a message
+   on this project — `push_tokens` has been empty since the fanout was first
+   scheduled, no `apps/mobile/google-services.json` exists in this checkout, and
+   the iOS APNs key is recorded as not done (`README.md`, "Turning on push").
+   The in-app Inbox screen that used to be the fallback was removed in #565. So
+   a `settlement_confirmed`, an `expense_disputed`, a `ghost_claim_approved` or a
+   trip nudge is written to `notifications`, marked `push_status = 'failed'`
+   terminally by the claim (no live token is a _decision_, not a retry), never
+   mailed (not on the email whitelist), and never seen.
+
+2. **The most obvious notification in an expense-splitting app does not exist.**
+   `expense_added`, `expense_edited`, `expense_deleted` and `you_owe` have copy,
+   a preference switch that governs them (`waves_pref_key_for_kind` →
+   `involvesMe`), a renderer and tests — and **no producer anywhere**. Nothing in
+   `expense-write`, `sync`, or any SQL function calls `waves_notify` for an
+   expense. The same is true of `settlement_initiated` (which has an _email
+   template_), `digest_daily` (also with a template), `group_invite_accepted` and
+   `ghost_claimed`. The complete producer list is eleven functions and is given
+   in §2.3.
+
+3. **Not one local notification is scheduled anywhere.** There is no
+   `scheduleNotificationAsync` call in the repo. `expo-notifications` is used
+   only to ask permission, fetch a push token, and route a tap
+   (`apps/mobile/src/lib/push.ts`). Everything proposed for the private ledger in
+   §4 is greenfield on the device — and, unlike the server half, needs no
+   credential, no migration and no cron.
+
+---
+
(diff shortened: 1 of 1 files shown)


dmadan86:
Decisions taken by the user on 2026-09-12, to fold into the document before implementation:

1. **Expense notifications go only to people the expense involves** — not everyone in the group. This matches the `involvesMe` pref switch that already exists.
2. **Phone-only accounts stay push-only.** Signup will not ask for an email address; those accounts simply never receive the summary or any other mail.
3. **Dunning stays manual.** The built-but-unfired `reminders.auto` path waits for quiet hours and a per-user timezone before it is turned on — a fixed hour in the group's timezone is not good enough for a member abroad.

Defaults accepted without further discussion: summary cadence Off/Weekly/Monthly with weekly the default and daily never offered; local money reminders name the thing but not the amount; 15-minute coalescing window for expense activity.

Still open: questions 2, 6, 9 and 10 (the `weeklyEmail` boolean migration, trip recap transport, what to do with the half-built `digest_daily`/`settlement_initiated`, and per-group granularity).

PR #793, opened 2026-09-13
Plan: the drafts hub and the rules engine

dmadan86:
A plan only — no feature code. `docs/plan-drafts-and-rules.md`, written against
the Splitkaro recording (a **Drafts** screen with *SMS Drafts*, *SMS Rules* and
*Offline Drafts* lanes, an Auto fetch toggle, an "Added from SMS" tag and a
"add 133 older transactions" banner) and read back against what this repo
already has.

## The permission verdict

> **Waves cannot read the SMS inbox and stay on Google Play.**

`READ_SMS` / `RECEIVE_SMS` are restricted permissions. The Permissions
Declaration Form takes a fixed list of use cases — default SMS handler, default
phone handler, backup-and-restore, device automation, connected device,
enterprise management, caller ID, OEM — and expense tracking is not on it, with
no way to argue for a case that is not listed. The compliance playbook rates it
`GOOGLE-PERM-SMS-CALLLOG`, **critical**, because the enforcement is removal from
Play rather than a rejected update.

This repo already reached that conclusion. #114 shipped an Android inbox reader;
#241 deleted it and the `READ_SMS` declaration with it; `waves-adr.md:18` says
*"pasting proved a better trade than holding an inbox-read permission"*. Today
`app.json` declares `"permissions": []`.

So the Auto fetch toggle, the "134 not added" count and the backfill banner are
all off the table. Everything downstream of a parsed transaction is not. iOS has
no SMS API at any tier, so anything built on the inbox would be an Android-only
app anyway.

The three Play-safe ways a transaction can arrive: **pasted**, **shared in** via
a share-sheet target, or **parsed out of an emailed statement**. Two of the three
work on iPhone.

One housekeeping find: `react-native-get-sms-android` is still a declared
dependency with a patch, imported by nothing. Its library manifest declares no
permissions (only its bundled example app does), so it is not leaking `READ_SMS`
today — but it should go, along with the dormant `sms_inbox_read` flag row, so
there is no switch to find later.

## What already exists vs what is new

Most of the drafts half is built and dark:

- `packages/core/src/sms/parse.ts` — the bank-SMS parser, 337 lines, tested,
  imported by nothing in the app.
- `packages/core/src/import/email.ts` — a **card-statement / booking-email
  parser** that landed in #397 and has **never had a UI**. Multi-row, shares the
  SMS dedupe key, guesses a category, needs no permission, works on iPhone. This
  is the real backfill path.
- `packages/core/src/category/merchant.ts` — `normaliseMerchantName` /
  `merchantTokens` already turn `"POS UPI SWIGGY*ORDER 8842"` into
  `"swiggy order"`. That is exactly what a learned rule must remember.
- `captures` is already the SMS-drafts row — description, amount, currency,
  `raw_text`, `parsed` jsonb, `payment_method`, `target_group_id`, status
  open/assigned. `capture.assign` is already the "Add to" button, and
  `planCaptureAssign` is already a bulk assign.
- ~45 `smsImport.*` strings are translated into four locales and referenced by
  nothing — the paste screen can ship with no new strings.
- `expense_versions.source` (`'manual' | 'imported'`) and `p_source` on
  `waves_apply_expense` both exist; `expenseWrite.ts:219` records that no client
  has ever set them.

Genuinely new: **the rules engine**. Nothing rule-shaped exists.

## The recommendation on the hub

**Do not build the three-lane hub.** Three lanes, three right homes:

1. **SMS/email drafts fold into `/captures`** — it is the drafts screen already.
   It gains a source filter (`SegmentedTabs`, not a left rail — RTL), a
   Waiting/Added pair, a provenance line, and free multi-select assign. A second
   list of the same rows under a different name is the two-inboxes mistake #565
   already undid.
2. **Rules get their own screen**, hung off the inbox — not under `personal/`,
   which is behind `PersonalGuard` and would demand a fingerprint to edit where
   your Swiggy orders go.
3. **"Offline Drafts" becomes "Waiting to send"**, a list under `settings/sync`
   — and #786 is what earns it. Since that PR a phone keeps unsent work for 30
   days after an involuntary session loss, and the only surface is a header
   glyph that renders *nothing* when idle and names only the head stuck item
   otherwise.

Also: **no default rule.** The competitor needs a "My Personal" holding group
because its drafts have to be somewhere; a Waves capture is nowhere by
construction until somebody assigns it.

## Rules: where it runs, and where it lives

Every rule is evaluated **on the device**, as a pure function in
`packages/core`. `docs/notifications-and-email-strategy.md:46` already classifies
captures as local, and a rule over the personal ledger could not run anywhere
else — the sync edge validates `recordKind` and stores `data` verbatim. Nothing
here contradicts either recent doc.

Storage: a new owner-scoped `automation_rules` table with a `<owner>:rules`
mirror scope, following `group_pins` (#789) exactly — that PR is the template
and it was ~880 lines. A new table needs its Prisma model in the same commit or
`db:drift` fails. The cheaper alternative (a fifth `record_kind` on
`personal_records`, three edits, no new table) is discussed and rejected: a
routing rule is not a money record.

Learning is designed to never surprise: what is remembered is the **normalised
merchant token**, a rule is **offered** after a manual move rather than acquired
silently, every rule-placed row says *"by a rule"*, moving it elsewhere rewrites
the target, and three corrections disable the rule instead of rewriting it
again. Auto-posting into a shared group is off by default, because a group
expense is a claim against other people's money.

## Phasing

Nine PRs, each shipping something true. 0 remove the dead dependency · 1 paste a
message · 2 the inbox grows up · 3 Waiting to send · 4 rules stored and edited
(**deploy-gated**) · 5 rules that learn · 6 paste a statement · 7 share-sheet
target (native, not OTA) · 8 `source` on the ledger (**deploy-gated**). 1–3 run
in parallel with 4–6.

## Open questions

1. **When is a rule learned** — ask on the first manual move, or learn silently
   on the second and say so afterwards?
2. **Is "Added from SMS" private to the author**, or do the other group members
   see it?
3. **Plaintext rules or sealed ones?** A readable table tells the server you
   route CRED Club to your flat.
4. **Does the email/statement lane matter?** It is already written, needs no
   permission, works on iPhone, and is the only real backfill — but it is the
   least like the recording.
5. **"Waiting to send": a full screen, or one expandable row** inside
   `settings/sync`?
6. **Should `ignore()` ship in the first rules PR?** Credit-card bill payments
   are debits that are not expenses — but it is the one action that drops a
   draft without a person seeing it.
7. **Is any of this paid-tier?**


dmadan86:
Decisions taken by the user on 2026-09-13, to fold into the plan before implementation:

1. **The email / card-statement lane ships.** It is the iPhone answer and the only honest backfill — paste or forward a statement, extract every row at once, no permission on either platform. The parser already exists (`packages/core/src/import/email.ts`, #397) and needs a screen, not an engine. This raises phase 6 in priority rather than leaving it last.
2. **A rule is learned by asking on the first manual move**, never silently on the second. Move CRED Club to Flat once and it asks "Always put CRED Club in Flat?" — nothing is remembered without an explicit yes, so the app can never surprise somebody with routing they did not agree to.
3. **Rules are sealed on the device** before they sync, the same way the private Me ledger is — the server stores them and cannot read them. A plaintext rules table would tell the server which merchants somebody uses and who they live with. This is consistent with the plan's own finding that rules must be evaluated on the device anyway, so nothing is lost by sealing them.

Defaults accepted without further discussion, unless the implementation finds a reason to revisit: no default rule (a Waves capture is nowhere by construction); "Waiting to send" as one expandable row under `settings/sync` rather than a full screen; `ignore()` held back from the first rules PR, since it is the one action that drops a draft unseen; provenance ("Added from SMS") private to the author; and none of this behind the paid tier for now.

Unchanged and worth restating: the Android SMS-inbox read stays off the table — #114 shipped it, #241 removed it, and the ADR's "pasting proved a better trade than holding an inbox-read permission" still holds. Phase 0 (deleting the dead `react-native-get-sms-android` dependency and the dormant `sms_inbox_read` flag) should go in regardless of whether the rest is built.

PR #797, opened 2026-09-13
Research: what the SMS parser actually reads, measured against 25 real message shapes

dmadan86:
Answers two questions that were asked together but are different: **how a message can reach the parser**, and **what the parser makes of it**. The second was measured, not estimated — `parseSms` run against 25 bodies written the way real institutions write them. The corpus is committed so a fix is measured against the same set rather than a new one that happens to pass.

## 20 of 25

The five failures are the largest banks in India, not edge cases:

| case | result | cause |
| --- | --- | --- |
| **SBI UPI** | dropped | `AMOUNT` needs a currency token; SBI writes `debited by 150.0` |
| **ICICI account debit** | dropped | says `debited` *and* `credited` (of the payee); `debit === credit` bails |
| **Credit card spend** | **booked as income** | `CREDIT_WORDS` matches *credit* inside `Credit Card`; no debit verb matches `used for a transaction of` |
| **Refund** | dropped | `CREDIT_WORDS` has `refund`, message says `refunded` |
| **`350.00 INR`** | dropped | `AMOUNT` only matches a marker before the number |

**The credit-card one is the only one that writes something wrong.** The others lose a transaction the person can paste again; this one records a ₹2,500 purchase as ₹2,500 *received*, and everything downstream trusts `direction`.

## Six more parse but lie about the merchant

`MERCHANT` has no `i` flag, so HDFC's `To SWIGGY` never matches. The capture requires an uppercase first character, so `at bigbasket` never matches. `to VPA swiggy@icici` yields the literal word `VPA`. `credited to A/c` yields `A`.

Those last two score **confidence 1.0** — because confidence counts fields *found*, never fields *plausible*, and `SMS_LOW_CONFIDENCE` gates pre-selection on it. A row reading `VPA` that is pre-ticked is worse than a row with no merchant at all.

## The rejection side is fine

All six negatives refused, including an OTP quoting a real amount from the right sender — the false positive that would cost the most.

## Two findings about the source, not the parser

**RCS.** Banks are migrating alerts to RCS in India (Jio, Airtel and Vi all in production in 2026). `READ_SMS` cannot see an RCS message — different store, no third-party read API. Every bank that migrates is a bank whose transactions quietly stop appearing, with no error the person can see.

**Small UPI already stopped.** HDFC no longer sends SMS for UPI under ₹100 sent / ₹500 received — the chai and the auto, which is the wrong end of the range for a splitting app.

Neither kills the feature. Both mean the paste and statement lanes are not merely the iPhone consolation prize, and that the copy should promise to catch *bank messages*, not to catch *spending*.

## Unused accuracy win

`parseSms` takes a `sender` and does nothing with it. India's DLT regime makes every commercial sender a registered 6-character header (`AD-HDFCBK` = Airtel Delhi + `HDFCBK`), so a bank allowlist is a better first filter than a regex over the whole inbox — and a better answer to "why do you need my inbox". Strip the operator prefix first: the same bank arrives as `AD-`, `VM-` or `JD-` by carrier and circle.

## Scope

Docs only, no parser change. Recommendations are ordered; the first four are small self-contained changes to a pure function with an existing suite, and worth doing before anyone tests against a real inbox — first impressions of this feature are entirely its hit rate.

Limits are stated in §5: no real inbox was read, the corpus is India-heavy, and RCS displacement is from public reporting rather than measurement.

## Summary by CodeRabbit

* **Documentation**
  * Added a measurement corpus covering 25 synthetic bank SMS messages, expected classifications, parser results, and scoring.
  * Added coverage documentation describing supported message channels, message families, parser outcomes, and known extraction and confidence-scoring issues.
  * Documented reproducibility details, measurement limitations, rejection behavior, and prioritized areas for improvement.

### DIFF
--- a/docs/sms-parser-corpus.md
+++ b/docs/sms-parser-corpus.md
@@ -1,0 +1,49 @@
+# The SMS corpus these results were measured against
+
+25 bodies written to match documented institution formats — **not captured from
+any real device or inbox**. Kept so that a parser change can be measured against
+the same set rather than against a new one that happens to pass. See
+[sms-parser-coverage.md](sms-parser-coverage.md) for the results.
+
+`null` means the parser must reject it. Those six matter most: a wrong rejection
+loses a transaction the person can paste again, but a wrong acceptance writes a
+ledger entry nobody made.
+
+Result column records behaviour as measured on `packages/core/src/sms/parse.ts`
+at commit `01b5f678`, September 2026.
+
+| #   | case                                    | expected | result               | body                                                                                                                        |
+| --- | --------------------------------------- | -------- | -------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| 1   | SBI UPI (no currency token)             | debit    | **dropped**          | `Dear UPI user A/C X1234 debited by 150.0 on date 12Sep26 trf to SWIGGY Refno 526012345678. If not u? call 1800111109 -SBI` |
+| 2   | HDFC UPI sent                           | debit    | ok, no merchant      | `Sent Rs.245.00 From HDFC Bank A/C x1234 To SWIGGY On 12/09/26 Ref 526012345678 Not You? Call 18002586161`                  |
+| 3   | ICICI acct debit                        | debit    | **dropped**          | `ICICI Bank Acct XX123 debited for Rs 500.00 on 12-Sep-26; SWIGGY credited. UPI:526012345678. Call 18002662 for dispute.`   |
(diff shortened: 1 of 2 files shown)


dmadan86:
Superseded by #799, which cherry-picked this research (`f29b2047`) and then extended it: the corpus keeps cases 1–25 verbatim and adds 26–61 for the international hazards, and the coverage document now carries before/after measurements from the same harness rather than a single reading.

Both files are on `main` as of #796. Closing so the same two documents do not arrive twice by different routes.

Its findings all landed: the five India failures and every merchant bug are fixed in #799, the display half in #798. The one recommendation deliberately **not** taken is the sender allowlist as an inbox pre-filter — an allowlist written against Indian DLT headers would have silently dropped every European bank. Only the confidence-hint half is wired.

PR #804, opened 2026-09-13
Bank messages leave one account at a time

dmadan86:
## Summary
- add owner-scoped bank-message cleanup for retained/local private data wipes
- preserve whole-store sign-out cleanup and checkpoint the SMS WAL
- cover user/rider/traveller/financer shared-device cleanup boundaries

## Validation
- pnpm --filter @waves/mobile exec vitest run test/smsMessageStore.test.ts test/smsInbox.test.ts
- pnpm --filter @waves/mobile exec vitest run test/smsMessageStore.test.ts test/smsInbox.test.ts test/smsScan.test.ts test/smsAutoRead.test.ts test/smsCaptureId.test.ts
- pnpm --filter @waves/mobile run typecheck
- pnpm exec prettier --check apps/mobile/src/lib/smsMessageStore.ts apps/mobile/src/lib/smsMessageStore.web.ts apps/mobile/src/sync/provider.tsx apps/mobile/test/smsMessageStore.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Sign-out cleanup now removes only the departing account’s stored bank messages, preserving data for other accounts.
  * Local message storage is fully cleaned up during complete sign-out, including temporary database data.


dmadan86:
Reviewed against the key lifecycle. **Recommend not merging as-is** — the scoping is correct in isolation but incoherent with what happens three lines earlier on the same path.

## The problem

`clearLocalPrivateData` is only reached for `SessionEnd.SignedOut`, and `destroyedBy` gives that scope `key: true`:

```ts
case SessionEnd.SignedOut:
  return { mirror: true, unsent: true, credentials: true, cache: true, key: true };
```

That runs `syncEngine.clear()` → `store.reset()` → `destroyKey()`. And the key is **one per device**, not one per account — a single `SecureStore` entry (`rowCipher.DEK_STORE_KEY`).

So after this change, signing out of account A:

1. deletes A's rows (correct),
2. leaves B's rows in the table (the intent of the PR),
3. and destroys the only key that can open them.

B's bank messages end up neither removed from the phone nor readable by anybody, ever. That is strictly worse than deleting them: the module's promise — "signing out takes your messages off this phone" — becomes false, and nothing is gained in exchange, because there is no key left to recover them with.

## The scenario the PR is written against

A user, rider, traveller and financer "one after another" on a shared device can't actually produce two live owners' rows. There is no account switcher, and every session change runs `endSession`, which wipes. The only way A's rows outlive A is `SessionEnd.Lost` (a revoked token) — and that path sets `key: false`, doesn't call `clearLocalPrivateData` at all, and already leaves the SMS store untouched. Which is the behaviour this PR wants, already present, on the one path where it is coherent.

When B then signs out deliberately and hands the phone on, destroying A's leftovers is the *right* outcome, and it is exactly what the mirror does — `reset()` deletes every mirror row unscoped, for the same reason.

## Two more things

`forgetEverything` has no callers in `src` after this change. It survives only via this PR's own test, which is how dead code gets kept alive.

`test/smsMessageStore.test.ts` ends with a bare assertion outside any `it`:

```ts
expect(SmsKind.Expense).toBe('expense');
```

That runs at import and is not reported as a test — if it ever fails it surfaces as a collection error, not a failing case. Also `loadKey` is mocked as `{ bytes: new Uint8Array(32) }` where `MirrorKey` is a bare `Uint8Array`.

## What is worth keeping

`PRAGMA wal_checkpoint(TRUNCATE)` is a genuine improvement and I had missed it. `secure_delete = ON` zeroes freed pages in the main file but the WAL still holds pre-images, so without the checkpoint the bytes outlive the delete. I am landing that part separately.

If multi-account really is wanted on one device, the change that makes it true is a **per-owner key** in `rowCipher` — at which point owner-scoped deletion follows naturally and the mirror should get the same treatment. That is a much larger piece of work and worth its own design, not a four-file fix.

dmadan86:
Rebased onto `main` — the conflict is gone and the branch is mergeable.

The conflict itself was trivial: both this branch and #806 added `PRAGMA wal_checkpoint(TRUNCATE)` to `forgetEverything`. Kept main's copy, which carries the reasoning for why the WAL needs its own wipe. Full suite green on the rebased branch (1649, 125 files).

**My review is unchanged: I still recommend against merging the owner-scoping.** Re-checked the ordering while rebasing, since I am arguing this twice:

`clearLocalPrivateData(ownerId)` runs `syncEngine.clear()` → `store.reset()` → `destroyKey()`, and *then* `forgetBankMessages(ownerId)`. One device-wide key, destroyed before the scoped delete runs. So after this change, signing out of A leaves B's rows on disk sealed with a key that no longer exists.

There is also a corollary that removes the last of the benefit: because the key is destroyed on **every** deliberate sign-out, rows belonging to any earlier account are already unreadable by the time the next account starts. Coexistence only survives the `SessionEnd.Lost` path — which sets `key: false`, never calls this function, and already leaves the store alone.

So the two outcomes are:

- **today:** table dropped, space reclaimed, "your messages are off this phone" literally true;
- **with this change:** table retains unopenable ciphertext for an account nobody can restore access to.

Nothing is gained and the promise in the module header stops being true. If multi-account on one device is genuinely wanted, the change that makes it work is a **per-owner key** in `rowCipher` — and the mirror needs the same treatment, since it is wiped unscoped for exactly the same reason. That is a design worth doing properly rather than a four-file fix.

Happy to be overruled — it is one line to flip the import in `sync/provider.tsx` back. Left unmerged for you to call.

PR #805, opened 2026-09-13
Review asks when a shop points two ways

dmadan86:
## Summary
- make Review merchant destination inference require a uniquely busiest group
- add a user/rider/traveller/financer tie regression so evenly split shop history asks instead of guessing

## Validation
- pnpm --filter @waves/mobile exec vitest run test/merchantDestination.test.ts test/reviewFeed.test.ts
- pnpm --filter @waves/mobile run typecheck
- pnpm exec prettier --check apps/mobile/src/lib/merchantDestination.ts apps/mobile/test/merchantDestination.test.ts


dmadan86:
Reviewed. **Correct — merging.**

This is not a behaviour change so much as the code finally doing what the module's own header already promises:

> when a merchant's spending is split across several groups, the newest group only wins if it is also the one it went to most often. A shop you used twice on a trip and once at home keeps pointing at the trip; **a shop genuinely split down the middle asks instead.**

`newest.count >= busiest` let the newest group *win* a tie, so a shop split down the middle pointed at whichever group it happened to visit last — a coin flip rendered as knowledge, on a chip whose whole justification is that the swipe confirms something visible. The `busiestGroups === 1` clause is the missing half.

Checked the arithmetic rather than the description:

- `busiest` starts at 0 and every count is ≥ 1, so the first group always takes the `>` branch and `busiestGroups` can never be left at 0 — no off-by-one hiding in the initialiser.
- A single group is untouched: one entry, `busiestGroups === 1`, still answers.
- 2 vs 1 is untouched in both directions — newest-is-busiest still answers, newest-is-not still asks.
- `===` replacing `>=` is a no-op on its own (`busiest` is the maximum), so the diff has exactly one behavioural clause, which is the right size for this.
- `Map` iteration is insertion-ordered, so the tie count is deterministic rather than dependent on hash order.

Full mobile suite green on the branch (1640).

One optional tidy, not blocking: the header bullet still reads "the one it went to most often" where the rule is now "the *only* one it went to most often". The sentence after it already covers ties, so it is not misleading — just a word looser than the code now is.

PR #853, opened 2026-09-17
The group screen gets its three faces

dmadan86:
Phase 2a of the web parity work: the group screen. The web showed one flat list of the last thirty expenses and four links; the phone has had three tabs, a month-cut ledger and a per-row answer to "what did this bill do to me" since the Mobbin pass.

## Three faces

**Expenses** — the ledger, cut into calendar months. Each row now ends in the reader's own stake: what they put in beyond their share, or their share of what somebody else put in. The direction is said in words beside the date (`you lent · 14 Sep`), so the row still reads for somebody who cannot tell the two money colours apart, and the group's total moves down to the subtitle where it is context rather than the answer. A bill between other people shows no figure at all — "not involved", which is a different sentence from "settled".

Deleted rows hide behind a switch rather than being dropped, and show struck through and dimmed when revealed. The ledger is append-only and being able to see that is the point.

**Balances** — every member's net, then the shortest set of payments that clears them. Both were already computed on this page and stacked below the feed.

**Activity** — the group's own trail. `groupActivity` is new on the client: `recentActivity` reads across every group, and asking for sixty rows spanning twenty groups to show three of them is the wrong shape of request.

## `myStake` moves into core

It lived in `apps/mobile/src/data/activity.ts`. Its own comment said it was shared "so the coloured figure on an expense row means the same thing wherever it is read" — and the web is now a second reader, so it moves to `@waves/core` beside the balances. The phone re-exports it from where it was, so no call site changes.

It arrived with no tests. It has seven now, chiefly the distinction that matters: `null` (a bill between other people) and `0n` (you paid exactly what you owed) are different answers, and folding them together would make every expense somebody is not part of read as one they are square on. Also that amounts are parsed as `bigint`, not `Number`.

## Smaller things

- The activity feed's verb emoji (🧾 ✏️ 🗑️ 💸 ✅ 🚩 🤝 👋 ✨) become stroked icons — the last emoji left in the web's chrome. `activity.ts` becomes `.tsx` for it.
- `SegmentedTabs` is a real tablist: arrows move between tabs and follow the reading direction, so the left arrow moves the way the eye does in Arabic; only the selected tab is in the tab order.
- At phone width the page's own primary action drops. It was the second "Add an expense" within a thumb's reach of the one in the top bar.

## Verification

`pnpm lint`, `tsc --noEmit`, `vitest` (56 web + 7 new core), `next build` — green. Mobile and core typecheck unchanged.

Rendered against the built stylesheet at 1440 and 390, light and dark, using a markup harness: the screen itself sits behind a session, so this was checked as markup rather than signed in. The duplicate-action problem above is what that check found.

## Next in this phase

Expense detail depth (comments, edit history, receipt viewer), then itemise, simplify, pending settlements and member detail.

## Summary by CodeRabbit

- **New Features**
  - Added Expenses, Balances, and Activity tabs to group pages.
  - Group expenses are organized by month, with personal balance status and deleted-expense indicators.
  - Added an option to show or hide deleted expenses.
  - Added group-specific activity feeds with clearer visual icons.
  - Added accessible keyboard navigation for tab controls.
- **Improvements**
  - Added translated labels and messages in English, Tamil, Hindi, and Arabic.
  - Improved mobile placement and styling of the Add Expense action.

<h2><a href="https://app.greptile.com/api/retrigger?id=65604883"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with no outstanding correctness or repository-rule findings.

<h3>Summary</h3>

The PR redesigns the group page around Expenses, Balances, and Activity tabs, adds a group-scoped activity request, presents expenses by month with each reader's stake, and moves stake calculation into the shared core package. Changes since the previous review key the detail component by group ID, fully resetting group-specific state during same-route navigation.
- Adds an accessible segmented tablist with keyboard and RTL-aware navigation.
- Adds deleted-expense visibility controls and month-based ledger ordering.
- Replaces activity emoji with consistent Lucide icons.
- Adds shared stake and ledger-grouping tests.
- The two previous findings are resolved: expense activity entries now link to their detail pages, and changing groups remounts the detail state.

<details open><summary>Diagram</summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart LR
  Route["/g/[groupId]"] --> Keyed["GroupDetail keyed by groupId"]
  Keyed --> CoreData["Group, members, expenses, settlements"]
  Keyed --> GroupActivity["Group-scoped activity"]
  CoreData --> Expenses["Expenses tab: monthly ledger + myStake"]
  CoreData --> Balances["Balances tab: nets + transfers"]
  GroupActivity --> Activity["Activity tab: group trail"]
  Expenses --> Detail["Expense detail route"]
  Activity --> Detail
```
</details>

<sub>Reviews (4) · Last reviewed commit: ["Key the group screen on the group"](https://github.com/dmadan86/waves/commit/a9b23e3fe64281a1abb9d23ae34587ec13b5f37d)</sub>


dmadan86:
`migrations, RLS policies and ledger invariants` went red on this branch and passed on a re-run. It is not this PR's, and it is worth someone's attention anyway.

**Why it isn't this PR's:** the diff touches no SQL, no migration and nothing under `packages/db` (`git diff --name-only main...HEAD` matches none of them). The same test passed on `main` at b31e24df, and passed here on the re-run with no change pushed between the two.

**Why it is worth a look:** the failure was `campaigns.test.ts > the holdout > does not hold the same people out of every campaign` — `expected 45 to be greater than 90`, where the test draws 300 people across two campaigns at 50% holdout each.

With independent buckets that figure should sit near 150. I simulated the same FNV-1a-mod-100 that `waves_bucket` implements over 400 trials of the same shape: median 149, minimum 120, and not one trial under the threshold of 90. So 45 is not the tail of the distribution this test is written against — two campaigns produced strongly correlated cohorts on that run, which is precisely the bug the test's own comment says it exists to catch ("Zero would mean one shared bucket, which is the bug").

I have not chased it further: it is a migration-shaped fix in code this PR does not touch, and it changes who lands in a holdout, so it is not mine to widen this PR with. Flagging rather than filing since whoever owns campaigns will know in one look whether 45 is explicable.

PR #913, opened 2026-09-19
THROWAWAY: Save an expense, opened the way the dashboard opens

dmadan86:
## Not for merging

Built to be looked at. The question was whether this screen should open on a saturated panel like Home does, and that's easier to answer by seeing it than by imagining it — the honest answer may well be no.

```
BEFORE                         AFTER
  ✕   🧾 Save an expense        ╔══════════════════════════╗
                                ║ 🧾 Save an expense     ✕ ║ ← wash runs under
   (−)  ₹0  INR⌄  (+)           ║                          ║   the status bar
   [+₹5][+₹10][+₹50]…           ║  (−)  ₹0  INR⌄   (+)     ║ ← white digits
                                ║  [+₹5][+₹10][+₹50]…      ║ ← translucent
  Receipt      [Add receipt]    ╚══════════════════════════╝
                                  Receipt    [Add receipt]
```

## It borrows, it doesn't copy

- **`ScreenHero`** is the same shell Home, a group, Review and Bank messages already open with — wash under the status bar, rounded bottom, glyph and name on the top row. Not a reimplementation of that panel; the panel itself.
- **`AmountField`'s `hero` size with `tone="onBrand"`** is a pairing the component already documents as *"the amount inside a coloured header bar"*. It was written for exactly this and **had no caller** until now.

## What a wash forces

The stepper discs and quick-amount chips go translucent white. On a gradient there is no muted surface to sit on — **a grey pill over indigo reads as a hole, not a control** — so they take the dim-white treatment every other hero's secondary controls already wear.

`Screen` goes to `edges={[]}` so the wash can reach the status bar instead of starting below it.

## The Save button

**It was already 56pt.** `size="lg"` is the height a primary action is meant to be, so "make it bigger" had less room than it sounds. This forces 64 purely so the difference is visible side by side. If 56 turns out to have been right, that one style prop is the whole revert.

## Reverting

Delete `AmountHeroPanel.tsx`, restore two lines in `capture.tsx`. `AmountHeader` is untouched and still serves add-expense, so **the two screens diverge while this branch lives** — which is itself worth looking at: if the wash is right here, add-expense should follow, and if it isn't, this goes.

1844 tests pass, typecheck and lint clean. **Not seen on a device** — and this one really wants to be, because a gradient behind a text input is exactly where contrast and keyboard behaviour surprise you.

https://claude.ai/code/session_0142pah41sBNEVSrrM1z4cXZ


dmadan86:
Reverted — the hero treatment is not going ahead. The branch is deleted; it was a throwaway built to be looked at, and it has been. Nothing on main ever carried it.

The two commits are recorded if this is ever worth revisiting: 31a4c1f3 (the hero panel) and 4060fe68 (the cherry-picked #912 fix, which lives on independently in its own PR).

PR #920, opened 2026-09-21
Home loses its coloured panel and gains one accent

dmadan86:
Following the GoodRx home screen you shared: white page, ink type, one accent colour used only behind glyphs. **Home only** — every other screen keeps the lavender page and the brand purple.

## The hero

It was a saturated gradient panel that crossfaded through four washes as the balance carousel moved, with a watermark glyph bled into the corner. It is now the page's own white. Two things follow, and they're why this is worth doing rather than a repaint:

- **The money gets its colour back.** On a coloured panel every figure was white, so "you are owed" and "you owe" were the same picture and only the caption told them apart — the exact fault [an earlier pass had to undo elsewhere in the app](https://github.com/dmadan86/waves/pull/191). On white they're the blue and the raspberry they are everywhere else. The month's spend stays ink: a total is not a direction, and colouring it raspberry would say it was a debt.
- **The greeting can be the page's title** — `title` rather than `heading`, now that nothing competes with it.

## The one element adopted wholesale

The quick-action discs take the launch yellow with the ink glyph on them. "Everything else" is the single inverted disc (ink filled, yellow glyph), so the way out of the row is findable without reading and without a second colour.

`accent` / `onAccent` join the theme for it: the icon's own yellow and the ink that reads on it, identical in both schemes because an accent that changed with the theme would stop being the thing people recognise on a home screen. It is decoration and never meaning — money keeps `positive`/`negative`, a warning keeps `warning`, and nothing sized like body text is drawn in it (yellow-on-white fails contrast at that size).

## Removed

The four gradient layers, the `SLIDE_STYLE` table that fed them, and `HeroBackdrop`. The carousel's shared scroll value stays — the pager pill still rides it, in the accent yellow instead of white.

## Not taken from the reference

Dotted row separators, outlined cards in place of shadowed ones, and the hand-drawn arrow doodle — you picked only the yellow discs, so cards keep their shadows and hairlines.

## Testing

`pnpm test:app` 1844 passed; mobile + `@waves/ui` typecheck and lint clean. **Not device-tested** — worth a look at the balance carousel in both themes, since the figures are coloured rather than white for the first time.

## Summary by CodeRabbit

- **Style**
  - Redesigned the dashboard hero with a cleaner white layout and improved text, balance, and status colors.
  - Updated balance figures to use clearer semantic colors for positive, negative, and neutral values.
  - Refreshed the dashboard pager indicator for better accent visibility.
  - Updated quick-action buttons with a consistent launch-yellow accent and improved icon contrast.
  - Added support for consistent accent colors across light and dark themes.


dmadan86:
Reverting this direction — Home keeps its coloured panel. Closing unmerged; the work stays on `design/home-goodrx` if we want to revisit it.

